PR #9, opened 2022-01-13
FIX: the image pull policy for ctlog, logserver, and signer

danpopnyc:
worked with @vaikas and saw that image was being pulled from external registry with the current pull policy.  removed from those yaml files.


vaikas:
Thanks so much! And thanks @mattmoor for debugging help!

mattmoor:
:party-matt:

PR #30, opened 2022-01-25
ADD: Mac client kind cluster hack script

danpopnyc:
edited default script to work on a mac client (disabled swapoff, and mapping to in memory etcd tmpfs

### DIFF
--- a/hack/setup-mac-kind.sh
+++ b/hack/setup-mac-kind.sh
@@ -1,0 +1,283 @@
+#!/usr/bin/env bash
+
+# Copyright 2021 Chainguard, Inc.
+# SPDX-License-Identifier: Apache-2.0
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+#if [[ -z "${GITHUB_WORKSPACE}" ]]; then
+#  echo "This script is expected to run in the context of GitHub Actions."
+#  exit 1
+#fi
+
+# Defaults
+K8S_VERSION="v1.21.x"
+KNATIVE_VERSION="1.1.0"
+REGISTRY_NAME="registry.local"
+REGISTRY_PORT="5000"
+CLUSTER_SUFFIX="cluster.local"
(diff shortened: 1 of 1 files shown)


hack/setup-mac-kind.sh
```diff
@@ -0,0 +1,283 @@
...
+#sudo swapoff -a
+#sudo rm -f /swapfile
+
+# Use in-memory storage to avoid etcd server timeouts.
+# https://kubernetes.slack.com/archives/CEKK1KTN2/p1615134111016300
+# https://github.com/kubernetes-sigs/kind/issues/845
+sudo mkdir -p /tmp/etcd
+#sudo mount -t tmpfs tmpfs /tmp/etcd
+
+curl -Lo ./kind "https://github.com/kubernetes-sigs/kind/releases/download/${KIND_VERSION}/kind-$(uname)-amd64"
```
vaikas, line 87: I was testing with arm64 when I was doing local testing, maybe change to arm64?

vaikas: or maybe change this to:
kind-$(uname)-$(uname -m)

danpopnyc: teh things i commented out related to the mapped drive to tmpfs which there is no concept of in MacOS i.e. https://superuser.com/questions/448585/is-there-an-implementation-of-tmpfs-for-mac-os-x

i removed the tmpfs mount and the mounted drives for the kind nodes

vaikas: Yeah, I had to make those changes as well :) However, is your mac intel or arm based?

danpopnyc: m1 (apple silcon)

vaikas: ok, well, if it worked for you as is, let's get it in, I'll retry it on my setup :)

PR #31, opened 2022-01-25
Update getting-started.md

danpopnyc:


getting-started.md
```diff
@@ -24,6 +24,12 @@ You should be able to install KinD and Knative bits with the following:
 ./hack/setup-kind.sh
 ```
 
+For a Mac,  please use this script (also please note that the airplay reciever uses the 5000 port and may need to be disabled details [here](https://developer.apple.com/forums/thread/682332)):
+
+```shell
+./hack/setup-mac-kind.sh
+```
```
vaikas, line 31: I also created a different release for arm64, at least on my Mac M1 I needed to run a different version of the mysql server, I cut a release on it last night, so maybe add a blurp about it. Were you running on arm or amd? If on amd, then I think we might need to have more smarts about the which version of the kind to run as well?
Anyhoo, for arm64, you need to run this version of the scaffolding release (release-arm.yaml)
https://github.com/vaikas/sigstore-scaffolding/releases/tag/v0.1.12-alpha

getting-started.md
```diff
@@ -24,6 +24,12 @@ You should be able to install KinD and Knative bits with the following:
 ./hack/setup-kind.sh
 ```
 
+For a Mac,  please use this script (also please note that the airplay reciever uses the 5000 port and may need to be disabled details [here](https://developer.apple.com/forums/thread/682332)):
```
vaikas, line 27: nit s/reciever/receiver/

Also, maybe say that if you want to you can change the port the registry listens if you don't want to disable the airplay receiver.

For this though, you have to download the release and manually change it, or run sed through it?

PR #45, opened 2022-02-14
Added flag for Trillian server addr

sabre1041:
Added support for specifying the location of the Trillian server as the backend for ctlog server


vaikas (approved):
Thank you!

PR #68, opened 2022-03-08
fix ko config

cpanato:
#### Summary
The ko config was pointing to a wrong path and had wrong flags, this PR fix that 

#### Ticket Link
n/a

#### Release Note

```release-note
NONE
```


vaikas (approved):
Thanks!

PR #88, opened 2022-04-04
Use apko as base image and add version information

cpanato:
#### Summary
- use apko v0.2.2 as base image
- output version information when the service start

#### Ticket Link
N/A

#### Release Note

```release-note
- use apko v0.2.2 as base image
- output version information when the service start
```


vaikas (approved):
Thank you so much!

PR #90, opened 2022-04-05
Use the go mod tidy -compat=1.17 in goreleaser hook

vaikas:
as per:
https://github.com/sigstore/scaffolding/runs/5824974530?check_suite_focus=true

Signed-off-by: Ville Aikas <vaikas@chainguard.dev>

#### Ticket Link

Fixes

#### Release Note

```release-note

```

PR #91, opened 2022-04-05
sync modules

cpanato:
#### Summary
follow up of #90 

cc @vaikas need this before we make a new release

#### Ticket Link
n/a

#### Release Note

```release-note
NONE
```


cpanato:
the reusable action tests is expected to fail

vaikas:
Is there some pre-check action that we're missing that we should have?

vaikas (approved):

cpanato:
> Is there some pre-check action that we're missing that we should have?

will think in some

PR #94, opened 2022-04-07
Add examples for signing and verifying an image, as well as

vaikas:
creating and verifying an attestation. Both with example
invocations so user can see what to expect.

#### Summary
Add examples for signing and verifying a local image as well as
creating and verifying an attestation. Both with fully cut&pasteable
command lines as well as example invocations so the user has an
idea what to expect.

#### Ticket Link

Fixes

#### Release Note

```release-note

```


cpanato (approved):
nice1!

PR #122, opened 2022-04-18
Add checks in setup-kind for existing steps

### LINKED ISSUE #64: Running setup-kind.sh keeps on creating registry entries in /etc/hosts
**Description**

Here we blindly always just write the entry for registry into the /etc/hosts.
https://github.com/sigstore/scaffolding/blob/main/hack/setup-kind.sh#L222

We should check if the entry already exists and not append if it does so.

eddiezane:
#### Summary
This PR fixes up some issue with running `hack/setup-kind.sh` multiple times.

#### Ticket Link

Fixes #64 

#### Release Note

```release-note
Fixes issues with setup-kind.sh on subsequent runs
```

### DIFF
--- a/hack/setup-kind.sh
+++ b/hack/setup-kind.sh
@@ -92,16 +92,22 @@ if [ ${RUNNING_ON_MAC} == "false" ]; then
   # See: https://kubernetes.slack.com/archives/CEKK1KTN2/p1600009955324200
   sudo swapoff -a
   sudo rm -f /swapfile
+
   # Use in-memory storage to avoid etcd server timeouts.
   # https://kubernetes.slack.com/archives/CEKK1KTN2/p1615134111016300
   # https://github.com/kubernetes-sigs/kind/issues/845
+  if mountpoint -q /tmp/etcd; then
+    sudo umount /tmp/etcd
+  fi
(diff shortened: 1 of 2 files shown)


cpanato:
i think we are not using this script anymore for ci testing, maybe only for local development

hack/setup-kind.sh
```diff
@@ -224,9 +230,11 @@ docker run -d --restart=always \
 # Connect the registry to the KinD network.
 docker network connect "kind" "$REGISTRY_NAME"
 
-# Make the $REGISTRY_NAME -> 127.0.0.1, to tell `ko` to publish to
-# local reigstry, even when pushing $REGISTRY_NAME:$REGISTRY_PORT/some/image
-sudo echo "127.0.0.1 $REGISTRY_NAME" | sudo tee -a /etc/hosts
+if ! grep -q "$REGISTRY_NAME" /etc/hosts; then
+  # Make the $REGISTRY_NAME -> 127.0.0.1, to tell `ko` to publish to
+  # local reigstry, even when pushing $REGISTRY_NAME:$REGISTRY_PORT/some/image
```
vaikas, line 235: nit registry.

vaikas (approved):
Thank you so much!!!

Issue #128, opened 2022-04-20
The wait logic for the e2e test is a bit flaky


k4leung4:
**Description**

Due to the scale to zero ability of knative serving, waiting on pods to be ready may some times fail inside the while loop as it scaled down while waiting for other pods to become ready.

 A short term fix is to set minscale to 1, to ensure it doesn't scale to zero.

PR #143, opened 2022-04-28
have always 1 pod running to avoid scale to 0 in ci

cpanato:
#### Summary
- have always 1 pod running to avoid scale to 0 in ci

#### Ticket Link

Fixes: https://github.com/sigstore/scaffolding/issues/128

#### Release Note

```release-note
NONE
```


k4leung4 (approved):

vaikas (approved):
Eventually we might set the Knative configmap so that cluster wide setting for minscale is 1, so we don't have to do them all, but since this is the last one let's goooo!

PR #147, opened 2022-05-02
Add Job for updating tree for sharding

priyawadhwa:
This marks a trillian merkle tree as "frozen" meaning that no more entries can be added.

we'll be running this job as part of the sharding process to freeze the existing log before creating a new active one!

### DIFF
--- a/cmd/trillian/createtree/main.go
+++ b/cmd/trillian/createtree/main.go
@@ -48,6 +48,7 @@ var (
 	displayName     = flag.String("display_name", "", "Display name of the new tree")
 	description     = flag.String("description", "", "Description of the new tree")
 	maxRootDuration = flag.Duration("max_root_duration", time.Hour, "Interval after which a new signed root is produced despite no submissions; zero means never")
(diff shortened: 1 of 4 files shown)


cmd/trillian/createtree/main.go
```diff
@@ -48,6 +48,7 @@ var (
 	displayName     = flag.String("display_name", "", "Display name of the new tree")
 	description     = flag.String("description", "", "Description of the new tree")
 	maxRootDuration = flag.Duration("max_root_duration", time.Hour, "Interval after which a new signed root is produced despite no submissions; zero means never")
+	force           = flag.Bool("force", false, "force create a new tree and update configmap")
```
cpanato, line 51: small nit

```suggestion
	force           = flag.Bool("force", false, "Force create a new tree and update configmap")
```

priyawadhwa: 👍🏽

cpanato (approved):

PR #158, opened 2022-05-13
Add in a read-only prober across Rekor and Fulcio API endpoints

priyawadhwa:
This prober performs all read requests across endpoints and exports data from prometheus. we can run it in the staging and production clusters to track metrics and set alerts.

### DIFF
--- a/config/prober/200-deployment.yaml
+++ b/config/prober/200-deployment.yaml
@@ -1,0 +1,41 @@
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  namespace: sigstore-prober
+  name: sigstore-prober
+  labels:
+    app: sigstore-prober
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      app: sigstore-prober
+  template:
+    metadata:
+      labels:
+        app: sigstore-prober
+      annotations:
+        prometheus.io/scrape: "true"
+        prometheus.io/path: /metrics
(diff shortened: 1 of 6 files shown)


config/prober/200-deployment.yaml
```diff
@@ -0,0 +1,27 @@
...
+      labels:
+        app: sigstore-prober
+      annotations:
+        prometheus.io/scrape: "true"
+        prometheus.io/path: /metrics
+        prometheus.io/port: "8080"
+    spec:
+      containers:
+      - name: sigstore-prober
+        image: ko://github.com/sigstore/scaffolding/cmd/prober
```
cpanato, line 25: maybe you can add some fine grain definition in the file https://github.com/sigstore/scaffolding/blob/main/.ko.yaml similar we have for the others

also do we need to release this image as well? or we build this adhoc?

priyawadhwa: thanks @cpanato, added in the definition!

so for this i was thinking we could release the image and then maybe package it into a helm chart so I could install it in staging and prod. WDYT?

cpanato: agree! then for that we need to add a rule in the make file and glue in the release, i can do that if you want as a follow up

priyawadhwa: that would be awesome, thank you :D

cpanato (approved):

PR #194, opened 2022-05-31
Add prober check for Fulcio write endpoint

priyawadhwa:

#### Ticket Link

Fixes

#### Release Note

```release-note

```


k4leung4 (approved):
lgtm

for future prs, would like to see the logic factored out, and unit tests added, and a e2e test using it. worried that prober will break and/or become incompatible with fulcio release.


would like to see unit tests,

cpanato:
can we create an issue to add unit tests and other things that maybe be needed?

PR #214, opened 2022-06-15
Add Rekor write endpoint to prober 

priyawadhwa:
Also Add HTTP Method to prometheus metrics so we can distinguish between reads and writes


cmd/prober/write.go
```diff
@@ -75,21 +85,89 @@ func fulcioWriteEndpoint(ctx context.Context) error {
...
 
+func rekorEntryRequest() ([]byte, error) {
+	payload := []byte(time.Now().String())
+	priv, err := cosign.GeneratePrivateKey()
+	if err != nil {
+		return nil, fmt.Errorf("generating keys: %w", err)
+	}
+	signer, err := signature.LoadECDSASignerVerifier(priv, crypto.SHA256)
+	if err != nil {
+		return nil, err
```
k4leung4, line 132: do we want to wrap the err to indicate exactly where it was failing?
same below

cmd/prober/write.go
```diff
@@ -75,21 +85,89 @@ func fulcioWriteEndpoint(ctx context.Context) error {
...
-	fmt.Println("Status code: ", statusCode)
-	fmt.Println("Latency: ", latency)
+	exportDataToPrometheus(resp, fulcioURL, endpoint, POST, latency)
+	return nil
+}
+
+// rekorWriteEndpoint tests the write endpoint for rekor, which is
+// /api/v1/log/entries and adds an entry to the log
+func rekorWriteEndpoint(ctx context.Context) error {
+	endpoint := "/api/v1/log/entries"
```
k4leung4, line 95: nit: make endpoints const?

.github/workflows/prober-test.yml
```diff
@@ -20,6 +20,7 @@ jobs:
     runs-on: ubuntu-latest
     permissions:
       id-token: write
+      packages: write
```
k4leung4, line 23: why do we need write on packages now?

priyawadhwa:
Trying to figure out why the prober test is failing -- looks like the prober isn't picking up ambient credentials from Github. Based on the [Github Docs](https://docs.github.com/en/actions/deployment/security-hardening-your-deployments/about-security-hardening-with-openid-connect#updating-your-actions-for-oidc) setting the `id-token: write` permission should be sufficient but that doesn't seem to be working (the `ACTIONS_ID_TOKEN_REQUEST_URL ` and `ACTIONS_ID_TOKEN_REQUEST_TOKEN ` env vars which should be set are not)

k4leung4:
> Trying to figure out why the prober test is failing -- looks like the prober isn't picking up ambient credentials from Github. Based on the [Github Docs](https://docs.github.com/en/actions/deployment/security-hardening-your-deployments/about-security-hardening-with-openid-connect#updating-your-actions-for-oidc) setting the `id-token: write` permission should be sufficient but that doesn't seem to be working (the `ACTIONS_ID_TOKEN_REQUEST_URL ` and `ACTIONS_ID_TOKEN_REQUEST_TOKEN ` env vars which should be set are not)

not entirely sure, but forks doesnt have access to the main repo secrets. you will have to merge it before it can gain access to the secret.

priyawadhwa:
> not entirely sure, but forks doesnt have access to the main repo secrets. you will have to merge it before it can gain access to the secret.

ah, that's probably it! guess the prober test can't be run as a precheck then, will have to switch it to run after merge.

cmd/prober/write.go
priyawadhwa: done!

cmd/prober/write.go
priyawadhwa: done

.github/workflows/prober-test.yml
priyawadhwa: copy/pasting leftover while debugging, it's removed now!

PR #215, opened 2022-06-15
add maintenance policy, avoid work hours for google maintenance

k4leung4:

#### Ticket Link

Fixes

#### Release Note

```release-note

```


cpanato (approved):
i would say we can add as a variable and default that to 6:00, but this is not a blocker

k4leung4:
ack, will do so in follow pr

PR #233, opened 2022-07-05
use workload identity for external secret instead of service key.

k4leung4:
#### Summary

ref:
https://external-secrets.io/v0.4.4/provider-google-secrets-manager/#using-service-accounts-directly

towards: https://github.com/sigstore/public-good-instance/issues/443

### DIFF
--- a/terraform/gcp/modules/external_secrets/external_secrets.tf
+++ b/terraform/gcp/modules/external_secrets/external_secrets.tf
@@ -15,10 +15,15 @@
  */
 
 
+locals {
+  namespace = "external-secrets"
+  k8s_sa    = "external-secrets"
+}
+
 // External-Secrets
 resource "helm_release" "external_secrets" {
   name             = "external-secrets"
-  namespace        = "external-secrets"
+  namespace        = local.namespace
   create_namespace = true
   chart            = "external-secrets"
   repository       = "https://charts.external-secrets.io"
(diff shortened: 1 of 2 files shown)


terraform/gcp/modules/external_secrets/external_secrets.tf
```diff
@@ -54,27 +66,6 @@ resource "google_service_account_key" "external_secrets_key" {
...
-  depends_on = [
-    google_service_account.external_secrets_sa,
-    google_service_account_key.external_secrets_key,
-    helm_release.external_secrets
-  ]
-}
-
 resource "kubectl_manifest" "secretstore_gcp_backend" {
   yaml_body = <<YAML
 apiVersion: external-secrets.io/v1alpha1
```
cpanato, line 71: maybe we can bump to `apiVersion: external-secrets.io/v1beta1` and use the new version

cpanato (approved):
lgtm

non blocking comment

terraform/gcp/modules/external_secrets/external_secrets.tf
k4leung4: we are still on 0.4.4, i dont think it supports v1beta1

though we should upgrade external secrets to 0.5x
filed https://github.com/sigstore/public-good-instance/issues/480

cpanato: i think we don't use anything that breaks, but i will double check

terraform/gcp/modules/external_secrets/external_secrets.tf
```diff
@@ -42,10 +47,17 @@
   depends_on = [google_service_account.external_secrets_sa]
 }
 
+resource "google_project_iam_member" "external_secrets_binding_token" {
+  project    = var.project_id
+  role       = "roles/iam.serviceAccountTokenCreator"
```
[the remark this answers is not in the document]
var-sdk: Post submit feedback @k4leung4 : can this be updated to grant the role on the specific service accounts(s) that need impersonating? This is a very powerful grant since the external-secrets SA can use access of any other service accounts in the project.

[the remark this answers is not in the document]
k4leung4: @var-sdk good point. created https://github.com/sigstore/scaffolding/pull/264

PR #249, opened 2022-07-21
Allow creating alerts with multiple notification channels

priyawadhwa:
This way we can send alerts to PD and Slack

### DIFF
--- a/terraform/gcp/modules/monitoring/dex/variables.tf
+++ b/terraform/gcp/modules/monitoring/dex/variables.tf
@@ -43,11 +43,11 @@ variable "dex_url" {
 }
 
 // Set-up for notification channel for alerting
-variable "notification_channel_id" {
-  type        = string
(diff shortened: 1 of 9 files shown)


terraform/gcp/modules/monitoring/dex/variables.tf
```diff
@@ -44,10 +44,10 @@ variable "dex_url" {
 
 // Set-up for notification channel for alerting
 variable "notification_channel_id" {
```
k4leung4, line 46: not sure if it is worthwhile to do or not, but rename to "notification_channel_ids"

and updating the description to indicate it takes more than one id?

same in other modules

priyawadhwa: done

k4leung4 (approved):
thanks

priyawadhwa:
@k4leung4 sorry had a lint error, could you re-approve?? thanks!!

k4leung4 (approved):

PR #262, opened 2022-08-02
Add a tuf server as well as repo management for tuf.

vaikas:
#### Summary
Starts fixing https://github.com/sigstore/sigstore/issues/544

We've been relying on ENV variables to target a custom sigstore installation. We want to move away from it, and first step is actually creating a custom tuf root. This adds support for it in scaffolding.
I'll follow up with changing the tests to use tuf, update documentation and github action for it.

Thank you @asraa for working together on this!

#### Release Note

* Add new service: tuf.tuf-system.svc that can serve the TUF root that gets created from the Fulcio cert, CT Log public key and from Rekor public key.

#### Documentation
Yes, documentation changes are necessary and I'll work on them as I update the e2e tests as well as update the release scripts.

PR #263, opened 2022-08-03
Refactor the github action, test with tuf root.

vaikas:
#### Summary
* Clean up and unify secret reconcile code.
* Add Rekor public key so we can have the 'oob' way of getting it. Will update rekor to have fileca (just like Fulcio) in a future PR so that we can use that.
* Modify e2e tests to use new tuf root
* Fix the e2e tests flow to work with 1.23, 1.24. But testing with 2[34] also introduced some oidc wonkiness with the k8s cluster because the oidc issuer on the kind cluster behaves differently, so had to work around that.
* Refactor the e2e tests because now we need to shuffle secrets between namespaces. This will be the basis for the new install-scaffold.
* Add new make targets for the above.
* Update README to reflect new reality.
* Address leftover comments from #262 except one (how to handle tuf server restart), but created issue: https://github.com/sigstore/scaffolding/issues/265 for it.


.github/workflows/fulcio-rekor-kind.yaml
```diff
@@ -122,25 +118,84 @@ jobs:
...
+      timeout-minutes: 5
+      run: |
+        make ko-apply-tuf
+
+        kubectl -n ctlog-system get secrets ctlog-public-key -oyaml | sed 's/namespace: .*/namespace: tuf-system/' | kubectl apply -f -
+        kubectl -n fulcio-system get secrets fulcio-pub-key -oyaml | sed 's/namespace: .*/namespace: tuf-system/' | kubectl apply -f -
+        kubectl -n rekor-system get secrets rekor-pub-key -oyaml | sed 's/namespace: .*/namespace: tuf-system/' | kubectl apply -f -
+
+        # Make sure the tuf jobs complete
+        kubectl wait --timeout 2m -n tuf-system --for=condition=Complete jobs --all
```
hectorj2f, line 188: Suggestion/question: Shouldn't the timeout be 4m ?

hectorj2f (approved):

.github/workflows/fulcio-rekor-kind.yaml
vaikas: This one seemed to consistently be quick, but for consistency, I'll change it :)

k4leung4 (approved):

PR #264, opened 2022-08-03
remove token creator role for external secrets.

k4leung4:

#### Summary

remove token creator role for external secret.
we should be able to leverage the existing service account for access.

ref:
https://external-secrets.io/v0.4.4/provider-google-secrets-manager/#using-pod-based-workload-identity
https://github.com/sigstore/public-good-instance/blob/main/terraform/environments/staging/helm-charts-values/external-secrets.yaml#L15

FYI: @var-sdk

### DIFF
--- a/terraform/gcp/modules/external_secrets/external_secrets.tf
+++ b/terraform/gcp/modules/external_secrets/external_secrets.tf
@@ -47,13 +47,6 @@ resource "google_project_iam_member" "external_secrets_binding" {
   depends_on = [google_service_account.external_secrets_sa]
 }
 
-resource "google_project_iam_member" "external_secrets_binding_token" {
-  project    = var.project_id
-  role       = "roles/iam.serviceAccountTokenCreator"
-  member     = "serviceAccount:${google_service_account.external_secrets_sa.email}"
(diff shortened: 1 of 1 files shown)


priyawadhwa (approved):

terraform/gcp/modules/external_secrets/external_secrets.tf
```diff
@@ -76,14 +69,6 @@ spec:
   provider:
       gcpsm:
         projectID: "${var.project_id}"
-        auth:
```
var-sdk, line 79: Is this deletion is intentional? (I don't understand this part of the change so this may be a dumb question.)

k4leung4: this part is intentional, as we will be authenticating based on the service account attached to the external secrets deployment/pod instead of specifying the full authentication separately.

this is outlined here, https://external-secrets.io/v0.5.8/provider-google-secrets-manager/#using-pod-based-workload-identity

var-sdk: Great, thanks for the explanation.

var-sdk (approved):

PR #268, opened 2022-08-07
Break release into smaller chunks.

### LINKED ISSUE #252: Make the release artifacts more granular.
**Description**

While it's convenient to do a one fell swoop and launch all the jobs at the same time, it does make things slower because of the exponential back-off that jobs use. Since there are several jobs that get launched simultaneously and because previous jobs need to finish before the final ones do, by the time we get to later stage jobs, they are backing off into the minutes.

So, if we create the release artifacts along the lines of steps as in the README.md and then launch those and wait for them to finish before starting the next step, we could shave off (guesstimate, minutes?) off the startup time.

I think the natural break from README (trillian, rekor, ctlog, fulcio) might make sense and make things more granular. But I think even just pulling the trillian into it's own would be good.
There's this PR that might help with that:
https://github.com/google/trillian/pull/2754

So, if we get that merged in, then we can remove some bits from here. But, this is just keeping track of this thought, as well as depending on how long it takes me to get that in, might start some work here ahead 🤷

vaikas:
#### Summary
Fix #252 

#### Release Note
* Release artifact has been broken into smaller chunks release-[ctlog,fulcio,rekor,trillian,tuf].yaml files.
* Add hack/setup-scaffolding.sh as a release artifact that can be used to install the whole sigstore.

#### Documentation
After the release gets created, update getting-started.md with the updated instructions.


cpanato:
need a rebase

lgtm :) thanks for doing that

vaikas:
Thanks! Yeah, I built this on top of this:
https://github.com/sigstore/scaffolding/pull/267

So, kind of want that going in first, and then I'll rebase this on top of that.

vaikas:
Thanks @cpanato PTAL when you get a moment, little easier to review now :)

cpanato (approved):
thank you

PR #273, opened 2022-08-09
Remove latency alerts on uptime checks

priyawadhwa:
Context and discussion at https://github.com/sigstore/public-good-instance/issues/513


haydentherapper:
Cc @var-sdk

haydentherapper:
I’m not sure this is the best approach. I think the round trip monitoring and alerting in general should be removed, because it’s hard to accurately get latency info unless it is coming directly from the cluster. Too many variables before the request gets to the cluster and after it leaves it. 

I think per regions alerts are still valuable, because we are likely to see more traffic from certain regions, and that might reveal certain issues.

priyawadhwa:
>  I think the round trip monitoring and alerting in general should be removed

> I think per regions alerts are still valuable

Sorry if I misunderstood, but these seem like opposing suggestions. Do you think we should remove this alert or keep it the way it is?

haydentherapper:
Sorry, my thoughts are:
* Per region alerts are still useful if the source of the data is from the cluster. 
* Round trip alerts that measure from the start of the request aren’t as useful since they measure much more than the service latency.

var-sdk:
I agree with @haydentherapper. Our uptime checks are not useful for measuring latency for SLO (which I assume is where the 750ms comes from) for a couple of reasons:
1. Latency is dominated by round trips between regions.
2. The requests the probers are making are mostly trivial read requests.

We should measure and alert on latency using server metrics or metrics from probers in the same cluster or region (@haydentherapper's per region alerts.) If those alerts are firing, these uptime latency alerts will also fire and just add noise.

My suggestion would be to use these uptime alerts just for uptime so we are still aware of scenarios when the service is unreachable from outside the cluster/region if customers ask. There is a time out of 10s on those requests so we do get some extreme sanity checks on latency. We could optionally reduce that to something less than client default timeout if we want to get alerts before customers see a problem, but I think 10s is fine.

var-sdk (approved):

haydentherapper (approved):
Thanks for the change! Just to check, do we have alerting for metrics from the cluster? If not, let’s file an issue for it.

priyawadhwa:
>  Just to check, do we have alerting for metrics from the cluster? If not, let’s file an issue for it.

That's the one that is currently disabled and tracked in https://github.com/sigstore/public-good-instance/issues/380

priyawadhwa:
@k4leung4 or @cpanato PTAL whenever you have a chance :)

k4leung4 (approved):

PR #278, opened 2022-08-09
Roles should be applied directly to the mysql service account

priyawadhwa:
was applying the roles to the wrong type of service account, this should fix it


k4leung4 (approved):
sorry i didnt catch this last time.

thanks

PR #290, opened 2022-08-12
actually parse the release-version, update curl to fail properly

vaikas:

#### Summary
Previously the release-version was hard coded and not used :(
Make curl fail on not found so it gives a meaningful error message instead of 'Not' not found when trying to execute a failed downloaded script.


vaikas:
Perfetto, hit the flake with v0.4.2 again here:
https://github.com/sigstore/scaffolding/runs/7802191065?check_suite_focus=true

So, updated to v0.4.3 as I was planning to do.

cpanato (approved):

PR #315, opened 2022-08-23
Update alignment period and duration for api prober error codes alert

priyawadhwa:
Previously this alert was misconfigured to fire every time a non-200 error code was seen.
Instead, we want to track the % of non-200 codes over the past minute (alignment_period).
If the % is nonzero for >5 minutes (duration), then send the alert.


k4leung4 (commented):
can you add a comment explaining the condition to make it easier to understand?

something like "average of non-200 status code over 60s, and when greater than 0 for 300s, alert"

not sure if that is right or not.

dlorenc (approved):

dlorenc (approved):

k4leung4 (approved):
thanks

cpanato (approved):

PR #331, opened 2022-08-30
add initial security policy resources

bobcallaway:
Signed-off-by: Bob Callaway <bcallaway@google.com>


k4leung4 (approved):

vaikas (dismissed):

terraform/gcp/modules/rekor/security_policy.tf
```diff
@@ -0,0 +1,77 @@
...
+ */
+
+// Enable required services for this module
+resource "google_compute_security_policy" "rekor" {
+  name    = "rekor-service-security-policy"
+  project = var.project_id
+  type    = "CLOUD_ARMOR"
+
+  rule {
+    action   = "deny(502)"
```
var-sdk, line 24: Any reason why not 413?

bobcallaway: Unfortunately the only options allowed by GCP are: 

deny(403), deny(404), deny(429) deny(502)

var-sdk (dismissed):
Sorry for the slow review, I went down the rabbit hole of trying to understand how these fit together with GKE and our helm charts. The factoring across repos is hard to grok but I don't have any better suggestions.

terraform/gcp/modules/rekor/security_policy.tf
```diff
@@ -0,0 +1,77 @@
...
+    priority = "1"
+
+    match {
+      expr {
+        expression = "int(request.headers['content-length']) > 134217728"
+      }
+    }
+    description = "Block all incoming requests > 128MB"
```
haydentherapper, line 32: Will this affect any current callers sending alpine packages? Or is this limit already in place?

terraform/gcp/modules/fulcio/security_policy.tf
```diff
@@ -0,0 +1,66 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+// Enable required services for this module
+resource "google_compute_security_policy" "fulcio" {
+  name    = "fulcio-service-security-policy"
+  project = var.project_id
+  type    = "CLOUD_ARMOR"
+
```
haydentherapper, line 22: Can we add a content length restriction here too? Someone could construct a massive CSR

terraform/gcp/modules/rekor/security_policy.tf
bobcallaway: same as https://github.com/sigstore/public-good-instance/blob/5c50dcf1b55537685609453b751507d136a05f29/argocd/utilities/templates/scaffold.yaml#L250

haydentherapper (approved):

PR #332, opened 2022-09-01
Add oslogin terraform module for optionally configuring OSLogin per project.

var-sdk:
Work towards https://github.com/sigstore/public-good-instance/issues/376. OSLogin enables us to control access to VMs without requiring provisioning of ssh keys and has better audit properties. The module is off by default so this should be a no-op.

### DIFF
--- a/terraform/gcp/modules/sigstore/sigstore.tf
+++ b/terraform/gcp/modules/sigstore/sigstore.tf
@@ -225,3 +225,29 @@ module "project_roles" {
   project_id           = var.project_id
   iam_members_to_roles = var.iam_members_to_roles
 }
+
+// OSLogin configuration
+module "oslogin" {
+  source     = "../oslogin"
+  project_id = var.project_id
+
+  // Disable module entirely if oslogin is disabled
+  count = var.oslogin.enabled ? 1 : 0
+
+  // Grant OSLogin access to the bastion instance to the GHA
+  // SA for terraform access and to tunnel accessors.
+  instance_os_login_members = {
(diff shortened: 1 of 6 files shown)


terraform/gcp/modules/oslogin/general.auto.tfvars
```diff
@@ -0,0 +1,17 @@
+project_id = "projectsigstore-staging"
```
cpanato, line 1: Do we need this file?

var-sdk: Oops, no that was a test file. Thanks for catching. Removed.

terraform/gcp/modules/sigstore/sigstore.tf
```diff
@@ -225,3 +225,29 @@ module "project_roles" {
...
   iam_members_to_roles = var.iam_members_to_roles
 }
+
+// OSLogin configuration
+module "oslogin" {
+  source     = "../oslogin"
+  project_id = var.project_id
+
+  // Disable module entirely if oslogin is disabled
+  count = var.monitoring.enabled ? 1 : 0
```
priyawadhwa, line 235: this should be `var.oslogin.enabled`

var-sdk: Fixed.

priyawadhwa (approved):
lgtm!

cpanato (approved):

cpanato:
@priyawadhwa it is good to merge?

PR #334, opened 2022-09-02
Refactor the CTLog config code, add multi-fulcio support.

vaikas:

#### Summary
Addresses (does not fix): https://github.com/sigstore/fulcio/issues/589

Start refactoring the code that manages the CTLog configuration and add support for adding/removing trusted
Fulcio certs + make it testable :)

This one currently adds a library, that I then intend to use, but since I have a question for @haydentherapper wanted to
share earlier.


pkg/ctlog/config.go
```diff
@@ -0,0 +1,324 @@
...
+	// This file contains the private key for the CTLog
+	privateKeyFile = "/ctfe-keys/privkey.pem"
+)
+
+// CTLogConfig abstracts the proto munging to/from bytes suitable for working
+// with secrets / configmaps. Note that we keep fulcioCerts here though
+// technically they are not part of the config, however because we create a
+// secret/CM that we then mount, they need to be synced.
+type CTLogConfig struct {
+	PrivKey         *rsa.PrivateKey
```
haydentherapper, line 68: Can we use an ECDSA key instead? It'll produce smaller signatures over SCTs, and that will result in smaller certificates since SCTs are embedded in issued certs.

pkg/ctlog/config.go
```diff
@@ -0,0 +1,324 @@
...
+	LogID           int64
+	LogPrefix       string
+
+	// Address of the gRPC Trillian Admin Server (host:port)
+	TrillianServerAddr string
+
+	// FulcioCerts contains one or more Root certificates for Fulcio.
+	// It may contain more than one if Fulcio key is rotated for example, so
+	// there will be a period of time when we allow both.
+	FulcioCerts [][]byte
```
haydentherapper, line 80: nit on the comment, it might also contain multiple if we choose to support admitting certificates from fulcio instances run by others.

pkg/ctlog/config.go
```diff
@@ -0,0 +1,324 @@
...
+		return nil, fmt.Errorf("unexpected number of Backends, want 1 got %d", len(multiConfig.Backends.Backend))
+	}
+	ret.TrillianServerAddr = multiConfig.Backends.GetBackend()[0].GetBackendSpec()
+
+	// Then we need to decode public key
+	pubPEM, _ := pem.Decode(public)
+	if pubPEM == nil {
+		return nil, fmt.Errorf("did not find valid public PEM data")
+	}
+	pubKey, err := x509.ParsePKCS1PublicKey(pubPEM.Bytes)
```
haydentherapper, line 166: when switching from rsa to ecdsa, this'll need to change to ParsePKIXPublicKey

pkg/ctlog/config.go
```diff
@@ -0,0 +1,324 @@
...
+	if !ok {
+		return nil, fmt.Errorf("Not a valid PEMKeyFile in proto")
+	}
+
+	privatePEMBlock, err := x509.DecryptPEMBlock(privPEM, []byte(pb.Password))
+	if err != nil {
+		return nil, fmt.Errorf("failed to decrypt private PEMKeyFile: %w", err)
+	}
+	ret.PrivKeyPassword = pb.Password
+	privKey, err := x509.ParsePKCS1PrivateKey(privatePEMBlock)
```
haydentherapper, line 190: This will need to change to ParsePKCS8PrivateKey

pkg/ctlog/config.go
```diff
@@ -0,0 +1,324 @@
...
+
+// MarshalSecrets returns a map suitable for creating a secret out of
+// containing the following keys:
+// private - CTLog private key, PEM encoded and encrypted with the password
+// public - CTLog public key, PEM encoded
+// fulcio-%d - For each fulcioCerts, contains one entry so we can support
+// multiple.
+func (c *CTLogConfig) marshalSecrets() (map[string][]byte, error) {
+	// Encode private key to PKCS#1 ASN.1 PEM.
+	block := &pem.Block{
```
haydentherapper, line 274: change marshalling to "PRIVATE KEY" and MarshalPkcs8PrivateKey, and below change to "PUBLIC KEY" and MarshalPKIXPublicKey

pkg/ctlog/config.go
```diff
@@ -0,0 +1,324 @@
...
+	)
+	data := map[string][]byte{
+		PrivateKey: privPEM,
+		PublicKey:  pubPEM,
+	}
+	for i, cert := range c.FulcioCerts {
+		fulcioKey := fmt.Sprintf("fulcio-%d", i)
+
+		// Fetch only root certificate from the chain
+		// TODO(Hayden): Is this what we should be adding, or the intermediates,
```
haydentherapper, line 303: This looks correct, it should only be the last certificate in the chain, the root.

pkg/ctlog/config_test.go
```diff
@@ -0,0 +1,298 @@
...
+		return nil, fmt.Errorf("GenerateKey failed: %w", err)
+	}
+	// Extract public component.
+	pub := key.Public()
+
+	serialNumber, err := rand.Int(rand.Reader, new(big.Int).SetInt64(math.MaxInt64))
+	if err != nil {
+		return nil, fmt.Errorf("failed to generate serial Number: %w", err)
+	}
+	rootCA := &x509.Certificate{
```
haydentherapper, line 213: See https://github.com/sigstore/rekor/blob/main/pkg/pki/x509/testutils/cert_test_utils.go, I created some test utils for creating root certs that are more representative of Sigstore root certs.

haydentherapper (commented):
This is looking great!

One higher level question - Since we'll be designing this to support multiple CT log instances, are there any changes we should be making in this refactor now to start supporting that? For example, if there's any places that expect one CT log config, making it an array? From looking over this, I think no if this is just setting up a single CT log's config, but I just wanted to check.

vaikas:
Thanks @haydentherapper I addressed your comments. I'll now convert the scaffolding to use this to make sure it works with the CTLog, since previously we were using the RSA keys for it.
https://github.com/sigstore/scaffolding/blob/main/cmd/ctlog/createctconfig/main.go#L114

haydentherapper (approved):
this looks great!

priyawadhwa (approved):

vaikas:
As I'm playing with this @haydentherapper and in particular with the tests about adding some new Fulcio to a pre-existing CTLog, we will have to deal with the fact that there may be RSA keys out there, so I'm trying to sort that out. Turns out, not everything is 'green-field' :)

PR #340, opened 2022-09-08
Remove /api/v1/index/retrieve endpoint from prober checks since it's experimental

priyawadhwa:
ref https://github.com/sigstore/rekor/issues/1021

cc @bobcallaway 

### DIFF
--- a/terraform/gcp/modules/monitoring/prober/prober_alerts.tf
+++ b/terraform/gcp/modules/monitoring/prober/prober_alerts.tf
@@ -117,7 +117,7 @@ resource "google_monitoring_alert_policy" "prober_data_absent_alert" {
       }
 
       duration = "300s"
-      filter   = format("resource.type = \"prometheus_target\" AND metric.type = \"prometheus.googleapis.com/api_endpoint_latency/summary\" AND metric.labels.host = \"%s\"", each.key)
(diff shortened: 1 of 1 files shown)


haydentherapper:
I had brought this up a while ago, but can we differentiate between critical prober alerts that should result in pages and ones that should result in tickets? I think we should still have visibility into the search API, but it doesn't need to trigger a page.

priyawadhwa:
> can we differentiate between critical prober alerts that should result in pages and ones that should result in tickets

I'm not sure, I don't think GCP let's you differentiate between alerts so it would have to be done in PagerDuty.

For now I can keep the endpoint data but turn off the alert on this endpoint?

haydentherapper (approved):
Can we also add a comment in the prober config that it does not generate an alert?

haydentherapper (approved):

vaikas:
@priyawadhwa I think you might need to update the ko?
```
- uses: imjasonh/setup-ko@v0.6
   with:
       version: v0.12.0
```

Could've sworn there were some screwiness that resulted in badness and it looks like things have failed twice?

vaikas:
LOL :) You already did it :)

vaikas (approved):

PR #341, opened 2022-09-08
Add back support for RSA for legacy deployed CTLogs.

vaikas:

#### Summary
While I started playing around with testing the newly created config library, realized that we need to still be able to
deal with legacy configs (created with the RSA private key). So handle them both + tests for unmarshaling/round-tripping.


pkg/ctlog/config.go
```diff
@@ -194,17 +199,22 @@ func Unmarshal(ctx context.Context, in map[string][]byte) (*CTLogConfig, error)
...
+	if privPEM == nil {
+		return nil, fmt.Errorf("did not find valid private PEM data")
 	}
 	ret.PrivKeyPassword = pb.Password
-	privKey, err := x509.ParsePKCS8PrivateKey(privatePEMBlock)
+
+	privatePEMBlock, err := x509.DecryptPEMBlock(privPEM, []byte(pb.Password))
 	if err != nil {
-		return nil, fmt.Errorf("failed to parse private key PEM: %w", err)
+		return nil, fmt.Errorf("failed to decrypt private PEMKeyFile: %w", err)
```
haydentherapper, line 210: Does this need to not return in order to fall through to trying to decode pkcs8?

vaikas: Sorry, I don't understand what you mean. Even in the legacy cases, these were always encrypted but it was just that the keys created were just created as RSA keys. Does that make sense? I think this should always fail if decrypt fails because the keys handed to CTLog must be encrypted.

haydentherapper: Yep sorry, was thinking this was doing a different operation. This looks good!

haydentherapper (approved):

priyawadhwa (approved):

PR #345, opened 2022-09-09
do not use setup-kind action from the head.

vaikas:

#### Summary
Action used to create kind cluster was pointing at `main` which is currently busted, so revert this to a previous working version.


priyawadhwa (approved):

cpanato (approved):
there is any fix in the action we need to do?

vaikas:
@cpanato absolutely, I'm debugging it, but didn't want others to be stuck on it while I fumble around :)

Issue #129, opened 2022-04-20
Add flag to store config in secret to `createctconfig`


priyawadhwa:
Instead of a configmap

This would work well with https://github.com/sigstore/helm-charts/pull/142#pullrequestreview-947739496, which expects the config to exist in a Secret.

vaikas:
I created this 'reconciler' for secrets, so doing this should now be very straightforward.
https://github.com/sigstore/scaffolding/blob/main/pkg/secret/secret.go#L34

And to use it:
https://github.com/sigstore/scaffolding/blob/main/cmd/rekor/createsecret/main.go#L72

I'll take a look at doing this next week.

vaikas:
@priyawadhwa I know the original issue request was for storing this in the secret instead of configmap. But the more I think about this, I think we should move away from storing the CTLog config in the configmap as an option all together. WDYT?

priyawadhwa:
SGTM!

vaikas:
We now only use the secret, so the flag has been removed.

PR #352, opened 2022-09-14
Use ctlog.config for creating certs, add managecaroots job, tests.

### LINKED ISSUE #292: Add ability to handle multiple Fulcio certs in the createctconfig.
**Description**

As discussed in https://github.com/sigstore/fulcio/issues/589 we should be able to handle rotation of Fulcio certs properly. This requires being able to add multiple Fulcio certs to be trusted by the CTLog.
I guess you just have to add certs here, currently we just add one.

This is the file that CTLog reads:
https://github.com/sigstore/scaffolding/blob/main/cmd/ctlog/createctconfig/main.go#L127

And here we create the secret containing the Fulcio cert to trust:
https://github.com/sigstore/scaffolding/blob/main/cmd/ctlog/createctconfig/main.go#L196

Which gets mounted here:
https://github.com/sigstore/scaffolding/blob/main/config/ctlog/ctlog/300-ctlog.yaml#L45

So, the tricky bit is that we don't want to just be append only :) So, have to think about how to add enough, but not too many of the certs and TBD how to remove them from here, once old Fulcio's have been turned down.

vaikas:

#### Summary
* Rework the ctlog create config job to use the new pkg/ctlog package.
* Add managectroots job that can be used to add / delete Fulcio (or other) root certs for a given CTLog. Fixes #292 
* Add e2e tests for rotating Fulcio key. Adds a new one, verifies both are there, then removes another one.
* Always add the CTLog config into a secret. See #129 for discussion and motivation
* Create ecdsa key for CTLog by default as suggested by @haydentherapper 
* Rejigger the README to put things in the correct order.

Addresses but does not fix: https://github.com/sigstore/public-good-instance/issues/524

### DIFF
--- a/.github/workflows/add-remove-new-fulcio.yaml
+++ b/.github/workflows/add-remove-new-fulcio.yaml
@@ -165,6 +165,8 @@ jobs:
       run: |
         ko apply -BRf ./testdata/config/new-fulcio
         kubectl wait --timeout 5m -n fulcio-system --for=condition=Ready ksvc fulcio-new
+        NEW_FULCIO_URL=$(kubectl -n fulcio-system get ksvc fulcio-new -ojsonpath='{.status.url}')
+        echo "NEW_FULCIO_URL=$NEW_FULCIO_URL" >> $GITHUB_ENV
 
     - name: Add new Fulcio to ctlog as trusted Fulcio
       run: |
@@ -177,11 +179,41 @@ jobs:
         kubectl -n ctlog-system get pods -oname | xargs kubectl -n ctlog-system delete
         sleep 10
 
-    # TODO: Verify these properly, add the new Fulcio to TUF root, sign & verify
     - name: Dump the trusted certs
       run: |
         curl ${{ env.CTLOG_URL }}/sigstorescaffolding/ct/v1/get-roots | jq .certificates
 
+    - name: Verify both Fulcio certs are there
+      run: |
+        go run ./cmd/ctlog/verifyfulcio/main.go \
+        --ctlog-url ${{ env.CTLOG_URL }} \
+        --log-prefix sigstorescaffolding \
+        --fulcio-url ${{ env.FULCIO_URL }} \
+        --fulcio-url ${{ env.NEW_FULCIO_URL }}
+
+    - name: Remove old Fulcio from ctlog as trusted Fulcio
+      run: |
+        ko apply -BRf ./testdata/config/remove-old-fulcio
+        sleep 2
+        kubectl -n ctlog-system wait --for=condition=Complete --timeout=180s job/remove-fulcio
+
+    - name: Restart ctlog pod again to pick up the changes
+      run: |
+        kubectl -n ctlog-system get pods -oname | xargs kubectl -n ctlog-system delete
+        sleep 10
+
+    - name: Dump the trusted certs
+      run: |
+        curl ${{ env.CTLOG_URL }}/sigstorescaffolding/ct/v1/get-roots | jq .certificates
+
+    - name: Verify that only new Fulcio cert is there
+      run: |
+        go run ./cmd/ctlog/verifyfulcio/main.go \
+        --ctlog-url ${{ env.CTLOG_URL }} \
+        --log-prefix sigstorescaffolding \
+        --fulcio-url ${{ env.NEW_FULCIO_URL }}
+
+    # TODO: Add the new Fulcio to TUF root, sign & verify.
     - name: Collect diagnostics
       if: ${{ failure() }}
       uses: chainguard-dev/actions/kind-diag@main
--- a/.github/workflows/add-remove-new-fulcio.yaml
+++ b/.github/workflows/add-remove-new-fulcio.yaml
@@ -188,8 +188,8 @@ jobs:
         go run ./cmd/ctlog/verifyfulcio/main.go \
         --ctlog-url ${{ env.CTLOG_URL }} \
         --log-prefix sigstorescaffolding \
-        --fulcio-url ${{ env.FULCIO_URL }} \
-        --fulcio-url ${{ env.NEW_FULCIO_URL }}
+        --fulcio ${{ env.FULCIO_URL }} \
+        --fulcio ${{ env.NEW_FULCIO_URL }}
 
     - name: Remove old Fulcio from ctlog as trusted Fulcio
       run: |
@@ -211,7 +211,7 @@ jobs:
         go run ./cmd/ctlog/verifyfulcio/main.go \
         --ctlog-url ${{ env.CTLOG_URL }} \
         --log-prefix sigstorescaffolding \
-        --fulcio-url ${{ env.NEW_FULCIO_URL }}
+        --fulcio ${{ env.NEW_FULCIO_URL }}
 
     # TODO: Add the new Fulcio to TUF root, sign & verify.
     - name: Collect diagnostics
--- a/.github/workflows/add-remove-new-fulcio.yaml
+++ b/.github/workflows/add-remove-new-fulcio.yaml
@@ -1,0 +1,189 @@
+name: Key rotation for Fulcio E2E Tests
+
+on:
+  pull_request:
+    branches: [ main ]
+
+permissions: read-all
+
+defaults:
+  run:
+    shell: bash
+    working-directory: ./src/github.com/sigstore/scaffolding
+
+concurrency:
+  group: fulcio-key-rotation-${{ github.head_ref }}
+  cancel-in-progress: true
+
+jobs:
+  fulcio-key-rotation:
+    name: e2e tests for fulcio key rotation
+    runs-on: ubuntu-latest
+
+    strategy:
+      fail-fast: false # Keep running if one leg fails.
+      matrix:
+        k8s-version:
+        - v1.25.x
+
+        leg:
+        - fulcio-key-rotation
+
+        go-version:
+        - 1.18
+
+    env:
+      GOPATH: ${{ github.workspace }}
+      GO111MODULE: on
+      GOFLAGS: -ldflags=-s -ldflags=-w
+      KO_DOCKER_REPO: registry.local:5000/knative
+      KOCACHE: ~/ko
+      COSIGN_EXPERIMENTAL: true
+
+    steps:
+    - uses: chainguard-dev/actions/setup-mirror@main
+    # https://github.com/mvdan/github-actions-golang#how-do-i-set-up-caching-between-builds
+
+    - name: Set up Go
+      uses: actions/setup-go@268d8c0ca0432bb2cf416faae41297df9d262d7f # v3.3.0
+      with:
+        go-version: ${{ matrix.go-version }}
+        check-latest: true
+
+    - name: Check out our repo
+      uses: actions/checkout@2541b1294d2704b0964813337f33b291d3f8596b # v3.0.2
+      with:
+        path: ./src/github.com/sigstore/scaffolding
+
+    - uses: actions/cache@v3
+      with:
+        # In order:
+        # * Module download cache
+        # * Build cache (Linux)
+        path: |
+          ~/go/pkg/mod
+          ~/.cache/go-build
+          ${{ env.KOCACHE }}
+        key: ${{ runner.os }}-go-${{ matrix.go-version }}-${{ hashFiles('**/go.sum') }}
+        restore-keys: |
+          ${{ runner.os }}-go-${{ matrix.go-version }}-
+
+    - uses: imjasonh/setup-ko@v0.6
+      with:
+        version: v0.12.0
+
+    - uses: sigstore/cosign-installer@f3c664df7af409cb4873aa5068053ba9d61a57b6 # v2
+
+    - name: Setup Cluster
+      uses: chainguard-dev/actions/setup-kind@main
+      id: kind
+      with:
+        k8s-version: ${{ matrix.k8s-version }}
+        registry-authority: registry.local:5000
+        cluster-suffix: cluster.local
+
+    - name: Setup Knative
+      uses: chainguard-dev/actions/setup-knative@main
+      with:
+        version: "latest"
+        serving-features: >
+          {
+            "kubernetes.podspec-fieldref": "enabled"
+          }
+
+    - name: Create sample image
+      run: |
+        pushd $(mktemp -d)
+        go mod init example.com/demo
+        cat <<EOF > main.go
+        package main
+        import "fmt"
+        func main() {
+          fmt.Println("hello world")
+        }
+        EOF
+        demoimage=`ko publish -B example.com/demo`
+        echo "demoimage=$demoimage" >> $GITHUB_ENV
+        echo Created image $demoimage
+        popd
+
+    - name: Install scaffolding
+      run: |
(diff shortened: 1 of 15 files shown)


.github/workflows/add-remove-new-fulcio.yaml
```diff
@@ -0,0 +1,221 @@
...
+    - name: Dump the trusted certs
+      run: |
+        curl ${{ env.CTLOG_URL }}/sigstorescaffolding/ct/v1/get-roots | jq .certificates
+
+    - name: Verify that only new Fulcio cert is there
+      run: |
+        go run ./cmd/ctlog/verifyfulcio/main.go \
+        --ctlog-url ${{ env.CTLOG_URL }} \
+        --log-prefix sigstorescaffolding \
+        --fulcio ${{ env.NEW_FULCIO_URL }}
```
haydentherapper, line 214: Very nice job, love being able to see all of this work composed together

cmd/ctlog/createctconfig/main.go
```diff
@@ -57,6 +60,7 @@ var (
 	ctlogPrefix        = flag.String("log-prefix", "sigstorescaffolding", "Prefix to append to the url. This is basically the name of the log.")
 	fulcioURL          = flag.String("fulcio-url", "http://fulcio.fulcio-system.svc", "Where to fetch the fulcio Root CA from")
 	trillianServerAddr = flag.String("trillian-server", "log-server.trillian-system.svc:80", "Address of the gRPC Trillian Admin Server (host:port)")
+	keyType            = flag.String("keytype", "ecdsa", "Which private key to generate [rsa,ecdsa")
```
haydentherapper, line 63:
```suggestion
	keyType            = flag.String("keytype", "ecdsa", "Which private key to generate [rsa,ecdsa]")
```

haydentherapper, line 63: We could also support ed25519 at some point if we wanted to, but i know that sigstore-python doesn't support it currently, and i'm not sure if other clients do. When writing the architecture docs, we'll want to decide whether to only support ecdsa and rsa, or mandate that clients support ed25519 too. i'm not incredibly opinionated here - If all languages had as nice of a key abstraction as golang, I'd say support all three, but it's probably unnecessary

cmd/ctlog/createctconfig/main.go
```diff
@@ -15,23 +15,24 @@
 package main
```
haydentherapper, line 15: nit: Should this be renamed since it doesn't just create a ctconfig but also updates? at the very least, we should document the update behavior

.github/workflows/add-remove-new-fulcio.yaml
vaikas: Aw, thanks!!!

cmd/ctlog/createctconfig/main.go
vaikas: Ok, fine as a follow up?

cmd/ctlog/createctconfig/main.go
```diff
@@ -111,113 +119,81 @@ func main() {
...
 
-	nsSecret := clientset.CoreV1().Secrets(ns)
-	if err := secret.ReconcileSecret(ctx, *secretName, ns, data, nsSecret); err != nil {
+	// Finally add Fulcio to it, marshal and write out.
+	existingConfig.AddFulcioRoot(ctx, root.ChainPEM)
+	marshaled, err := existingConfig.MarshalConfig(ctx)
+	if err != nil {
+		log.Fatalf("Failed to marshal new configuration: %v", err)
+	}
+	// Take out the public / private key from the secret since we didn't mess
```
haydentherapper, line 187: I'm unclear on why these are removed from the config?

cmd/ctlog/createctconfig/main.go
vaikas: Yeah, thought about that for renaming, but that would also have touch up things like charts, so felt it best to leave as is. Regarding the documentation, ✅ will do.

cmd/ctlog/createctconfig/main.go
```diff
@@ -230,3 +206,33 @@ func mustMarshalAny(pb proto.Message) *anypb.Any {
...
+func createConfigWithKeys(ctx context.Context, keytype string) (*ctlog.CTLogConfig, error) {
+	var privKey crypto.PrivateKey
+	var err error
+	if *keyType == "rsa" {
+		privKey, err = rsa.GenerateKey(rand.Reader, bitSize)
+		if err != nil {
+			return nil, fmt.Errorf("Failed to generate Private RSA Key: %w", err)
+		}
+	} else {
+		privKey, err = ecdsa.GenerateKey(elliptic.P256(), rand.Reader)
```
haydentherapper, line 222: Like RSA's bit size, we should make the elliptic curve configurable. this will save us time in the future if we need stronger keys. we can just support p256, p384, and p521.

cmd/ctlog/managefulcio/main.go
```diff
@@ -0,0 +1,173 @@
...
+	bitSize   = 4096
+)
+
+var (
+	cmname           = flag.String("configmap", "ctlog-config", "Name of the configmap where the treeID lives. if configInSecret is false, ctlog config gets added here also.")
+	configInSecret   = flag.Bool("config-in-secret", false, "If set to true, fetch / update the ctlog configuration proto into a secret specified in ctlog-secrets under key 'config'.")
+	secretName       = flag.String("secret", "ctlog-secrets", "Name of the secret to fetch private key for CTLog.")
+	pubKeySecretName = flag.String("pubkeysecret", "ctlog-public-key", "Name of the secret to fetch public key from.")
+	fulcioURL        = flag.String("fulcio-url", "http://fulcio.fulcio-system.svc", "Where to fetch the fulcio Root CA from.")
+	operation        = flag.String("operation", "", "Operation to perform for the specified fulcio [add,remove]")
```
haydentherapper, line 50: I think this could also be used for non-fulcio certificates too - let's say we add in third-party CA certificates. can we update the documentation and name of the job (managectroots?) for this? might want to update `fulcioOp` too

cmd/ctlog/verifyfulcio/main.go
```diff
@@ -0,0 +1,136 @@
...
+		fulcioCerts, err := cryptoutils.UnmarshalCertificatesFromPEM(root.ChainPEM)
+		if err != nil {
+			logging.FromContext(ctx).Fatalf("Failed to unmarshal fulcio Root cert %s cert: %v", fulcio.String(), err)
+		}
+		fulcioRoot, err := cryptoutils.MarshalCertificateToPEM(fulcioCerts[len(fulcioCerts)-1])
+		if err != nil {
+			logging.FromContext(ctx).Fatalf("Failed to marshal fulcio Root cert %s cert: %w", fulcio.String(), err)
+		}
+		logging.FromContext(ctx).Infof("Got a root cert for fulcio Root cert %s", fulcio)
+		// Strip the certificate begin/end marker strings since CTLog doesn't
```
haydentherapper, line 106: Is this accurate? https://github.com/sigstore/public-good-instance/blob/main/config/ctfe/root.pem shows that we did specify PEM headers. Or is this just how we store them in the ctlog config, and then add headers back later on?

cmd/ctlog/verifyfulcio/main.go
```diff
@@ -0,0 +1,136 @@
...
+		fulcioRootPEM := []byte(base64.StdEncoding.EncodeToString(block.Bytes))
+		for j := range certs.Certificates {
+			logging.FromContext(ctx).Infof("Checking ctlog root cert %s", certs.Certificates[j])
+			if bytes.Compare(fulcioRootPEM, []byte(certs.Certificates[j])) == 0 {
+				logging.FromContext(ctx).Infof("Found a matching root cert for fulcio Root cert %s cert: %w", fulcio.String(), err)
+				certsFound[i] = true
+			}
+		}
+	}
+	// Check that all the expected roots were found.
```
haydentherapper, line 121: As mentioned above, if we're supporting third-party CAs, we would want to check that we've found all fulcio roots, but it might just be a subset of roots in the ctlog config.

cmd/ctlog/verifyfulcio/main.go
```diff
@@ -0,0 +1,136 @@
...
+	}
+	certs := CertResponse{}
+	if err := json.Unmarshal(body, &certs); err != nil {
+		logging.FromContext(ctx).Fatalf("Failed to unmarshal body from ctlog: %v", err)
+	}
+	for i, cert := range certs.Certificates {
+		logging.FromContext(ctx).Infof("Got back cert %d: %s", i, cert)
+	}
+
+	// Keep track of certs found. Same index as fulcioURLs
```
haydentherapper, line 88: Can we guarantee that the order remains the same? It might be safer to build a map from cert serial number (which should be unique) to cert.

pkg/ctlog/config.go
```diff
@@ -81,29 +83,54 @@ type CTLogConfig struct {
 	FulcioCerts [][]byte
 }
 
+func extractFulcioRoot(fulcioRoot []byte) ([]byte, error) {
```
haydentherapper, line 86: nit: `fulcioCertChain` for a more precise variable name

haydentherapper (changes requested):

cmd/ctlog/createctconfig/main.go
haydentherapper: SG, agreed on avoiding larger renamings mixed into this

cmd/ctlog/createctconfig/main.go
haydentherapper: Yep, I think leave this as a TODO and we can implement this after clients add support

cmd/ctlog/verifyfulcio/main.go
vaikas: This is how they are being served from the logs:
```
vaikas@villes-mbp scaffolding % curl https://ctfe.sigstore.dev/test/ct/v1/get-roots | jq .
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  1381  100  1381    0     0   4538      0 --:--:-- --:--:-- --:--:--  4681
{
  "certificates": [
    "MIIB+DCCAX6gAwIBAgITNVkDZoCiofPDsy7dfm6geLbuhzAKBggqhkjOPQQDAzAqMRUwEwYDVQQKEwxzaWdzdG9yZS5kZXYxETAPBgNVBAMTCHNpZ3N0b3JlMB4XDTIxMDMwNzAzMjAyOVoXDTMxMDIyMzAzMjAyOVowKjEVMBMGA1UEChMMc2lnc3RvcmUuZGV2MREwDwYDVQQDEwhzaWdzdG9yZTB2MBAGByqGSM49AgEGBSuBBAAiA2IABLSyA7Ii5k+pNO8ZEWY0ylemWDowOkNa3kL+GZE5Z5GWehL9/A9bRNA3RbrsZ5i0JcastaRL7Sp5fp/jD5dxqc/UdTVnlvS16an+2Yfswe/QuLolRUCrcOE2+2iA5+tzd6NmMGQwDgYDVR0PAQH/BAQDAgEGMBIGA1UdEwEB/wQIMAYBAf8CAQEwHQYDVR0OBBYEFMjFHQBBmiQpMlEk6w2uSu1KBtPsMB8GA1UdIwQYMBaAFMjFHQBBmiQpMlEk6w2uSu1KBtPsMAoGCCqGSM49BAMDA2gAMGUCMH8liWJfMui6vXXBhjDgY4MwslmN/TJxVe/83WrFomwmNf056y1X48F9c4m3a3ozXAIxAKjRay5/aj/jsKKGIkmQatjI8uupHr/+CxFvaJWmpYqNkLDGRU+9orzh5hI2RrcuaQ==",
    "MIIB9zCCAXygAwIBAgIUALZNAPFdxHPwjeDloDwyYChAO/4wCgYIKoZIzj0EAwMwKjEVMBMGA1UEChMMc2lnc3RvcmUuZGV2MREwDwYDVQQDEwhzaWdzdG9yZTAeFw0yMTEwMDcxMzU2NTlaFw0zMTEwMDUxMzU2NThaMCoxFTATBgNVBAoTDHNpZ3N0b3JlLmRldjERMA8GA1UEAxMIc2lnc3RvcmUwdjAQBgcqhkjOPQIBBgUrgQQAIgNiAAT7XeFT4rb3PQGwS4IajtLk3/OlnpgangaBclYpsYBr5i+4ynB07ceb3LP0OIOZdxexX69c5iVuyJRQ+Hz05yi+UF3uBWAlHpiS5sh0+H2GHE7SXrk1EC5m1Tr19L9gg92jYzBhMA4GA1UdDwEB/wQEAwIBBjAPBgNVHRMBAf8EBTADAQH/MB0GA1UdDgQWBBRYwB5fkUWlZql6zJChkyLQKsXF+jAfBgNVHSMEGDAWgBRYwB5fkUWlZql6zJChkyLQKsXF+jAKBggqhkjOPQQDAwNpADBmAjEAj1nHeXZp+13NWBNa+EDsDP8G1WWg1tCMWP/WHPqpaVo0jhsweNFZgSs0eE7wYI4qAjEA2WB9ot98sIkoF3vZYdd3/VtWB5b9TNMea7Ix/stJ5TfcLLeABLE4BNJOsQ4vnBHJ"
  ]
}
```

cmd/ctlog/createctconfig/main.go
vaikas: Ah, because the reconcilesecret only 'reconciles' the keys that are in this map. We pull them out of the map so that they are never modified by the run of this script.
https://github.com/sigstore/scaffolding/blob/main/pkg/secret/secret.go#L28

cmd/ctlog/verifyfulcio/main.go
vaikas: Thought we might need that, so thanks for that use case :) 

There's the --strict flag that we can then flip to false. For the tests, I wanted to make sure it was a precise match, but thought there would be other use cases.
https://github.com/sigstore/scaffolding/pull/352/files#diff-f3a6ec0f5f1577f486baea8af21fd6f52dc4a009ed193babaef0b758e09eb841R57

cmd/ctlog/verifyfulcio/main.go
vaikas: Throughout the run of the script, yes. But, let me double-check :)

vaikas: Yes, because we do bytewise comparison, so it's the same index.

vaikas:
Thanks for the review @haydentherapper I think I addressed all your comments, PTAL :)

priyawadhwa (approved):

cmd/ctlog/verifyfulcio/main.go
haydentherapper: Looks good!

cmd/ctlog/verifyfulcio/main.go
haydentherapper: Ah interesting, this is base64 encoded DER. I guess `pem.Decode` handles headerless PEM?

haydentherapper (commented):
Late LGTM, looks good!

PR #353, opened 2022-09-14
Fix prober alerts.

var-sdk:
The API doesn't accept the "!=~" shorthand like the UI.

### DIFF
--- a/terraform/gcp/modules/monitoring/prober/prober_alerts.tf
+++ b/terraform/gcp/modules/monitoring/prober/prober_alerts.tf
@@ -161,7 +161,7 @@ resource "google_monitoring_alert_policy" "prober_error_codes" {
 
       comparison      = "COMPARISON_GT"
       duration        = "300s"
-      filter          = "resource.type = \"prometheus_target\" AND metric.type = \"prometheus.googleapis.com/api_endpoint_latency_count/summary\" AND metric.labels.status_code !=~ \"20[0|1]\" AND metric.labels.endpoint != \"/api/v1/index/retrieve\""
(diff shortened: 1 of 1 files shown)


terraform/gcp/modules/monitoring/prober/prober_alerts.tf
```diff
@@ -161,7 +161,7 @@ resource "google_monitoring_alert_policy" "prober_error_codes" {
 
       comparison      = "COMPARISON_GT"
       duration        = "300s"
-      filter          = "resource.type = \"prometheus_target\" AND metric.type = \"prometheus.googleapis.com/api_endpoint_latency_count/summary\" AND metric.labels.status_code !=~ \"20[0|1]\" AND metric.labels.endpoint != \"/api/v1/index/retrieve\""
+      filter          = "resource.type = \"prometheus_target\" AND metric.type = \"prometheus.googleapis.com/api_endpoint_latency_count/summary\" AND metric.labels.status_code != monitoring.regex.full_match(\"20[0|1]\") AND metric.labels.endpoint != \"/api/v1/index/retrieve\""
```
k4leung4, line 164: i think the syntax is `!~` instead of `!=~`

full regex match works too

k4leung4: https://cloud.google.com/monitoring/mql/reference#lexical-elements

var-sdk: I copied from the UI, which uses !=~ instead of !~ and then converts to full regex match underneath. Either are probably fine - I just want to unbreak terraform.

k4leung4 (dismissed):

terraform/gcp/modules/monitoring/prober/prober_alerts.tf
k4leung4: thanks for the clarification

var-sdk: After checking the documentation, I switched to the more readable and robust "one_of". PTAL and merge if you approve.

k4leung4 (approved):
that is more readable.
thanks

PR #366, opened 2022-09-20
Stop testing way older versions. build/test with go1.19

vaikas:

#### Summary

* Stop testing unnecessary older versions.
* build/test with go1.19 since that's what our go.mod sez
* test with k8s v1.25.x

I think we'll need to also go change the required tests since the matrix is changing? @cpanato


cpanato:
Will update

cpanato (approved):

PR #368, opened 2022-09-22
refactor signing release images

cpanato:
#### Summary
the new version of `ko` output in the `imagerefs` file all the images generated, before was just the main one and now we need to sign a bunch of images, and that can cause some timeout 

this PR tries to address this by splitting each ko build in their one session of signing :)

test locally and that works without problems

failure: https://github.com/sigstore/scaffolding/actions/runs/3101190107/jobs/5024860282


scripts/sign-release-images.sh
```diff
@@ -0,0 +1,30 @@
...
+set -o errexit
+set -o nounset
+set -o pipefail
+
+: "${GIT_HASH:?Environment variable empty or not defined.}"
+: "${GIT_VERSION:?Environment variable empty or not defined.}"
+
+if [[ ! -f $ARTIFACT ]]; then
+    echo "$ARTIFACT not found"
+    exit 0
```
priyawadhwa, line 26: do we need to fail if the ARTIFACT isn't provided?

priyawadhwa (approved):

scripts/sign-release-images.sh
cpanato: opssy :)

cpanato: thanks for the catch! i was doing some tests and forgot to revert

priyawadhwa (approved):

cpanato:
merging

PR #375, opened 2022-09-23
Set User Agent header and add prometheus metrics for prober build info

cpanato:
#### Summary
- export version information to prometheus
- set user agent when doing the requests

example of the prometheus metric

```
# HELP sigstore_prober_build_info A metric with a constant '1' value labeled by version, revision, branch, and goversion from which sigstore_prober was built.
# TYPE sigstore_prober_build_info gauge
sigstore_prober_build_info{build_date="2022-09-21T21:16:41",goversion="go1.19.1",revision="df49efcde361792134e16e55df6fbed8090a5e77",version="v0.4.8-dirty"} 1
```

FIxes: https://github.com/sigstore/public-good-instance/issues/749

### DIFF
--- a/cmd/prober/write.go
+++ b/cmd/prober/write.go
@@ -80,6 +80,7 @@ func fulcioWriteEndpoint(ctx context.Context) error {
 	req.Header.Set("Authorization", "Bearer "+tok)
 	// Set the content-type to reflect we're sending JSON.
 	req.Header.Set("Content-Type", "application/json")
+	req.Header.Set("User-Agent", fmt.Sprintf("Sigstore_Scaffolding_Prober/%s", versionInfo.GitVersion))
 
 	t := time.Now()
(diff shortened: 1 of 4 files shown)


priyawadhwa (approved):

cmd/prober/write.go
```diff
@@ -109,6 +110,7 @@ func rekorWriteEndpoint(ctx context.Context) error {
 	}
 	// Set the content-type to reflect we're sending JSON.
 	req.Header.Set("Content-Type", "application/json")
+	req.Header.Set("User-Agent", fmt.Sprintf("Sigstore_Scaffolding_Prober/%s", versionInfo.GitVersion))
```
vaikas, line 113: super nit, only do this once above and reuse. But for this is fine.

vaikas (approved):
Thanks!

PR #377, opened 2022-09-23
add missing ko configs and update base image path

cpanato:
#### Summary
- add missing ko configs and update base image path


vaikas (approved):
Thank you!

PR #382, opened 2022-09-24
Write the repository into secret, test air-gap mode.

vaikas:

#### Summary

TUF server now writes the compressed repo to a secret along with 1.root.json. This way we can test things like air-gap modes, especially in policy-controller without env variables.
Add a test simulating air-gap mode where we bring in a filesystem based TUF repository.

Was tested with:
https://github.com/sigstore/cosign/compare/main...vaikas:cosign:air-gap?expand=1

But I uncommented it for now so we can get this in and then I can add cosign e2e tests there as well once
we cut a release here.

#### Release Note

 * TUF root writes gzipped tar of the TUF repository, suitable for testing air-gap modes.

### DIFF
--- a/.github/workflows/fulcio-rekor-kind.yaml
+++ b/.github/workflows/fulcio-rekor-kind.yaml
@@ -156,6 +156,8 @@ jobs:
     - name: Initialize cosign with our custom tuf root and make root copy
       run: |
         kubectl -n tuf-system get secrets tuf-root -ojsonpath='{.data.root}' | base64 -d > ./root.json
+        # Also grab the compressed repository for airgap testing.
+        kubectl -n tuf-system get secrets tuf-root -ojsonpath='{.data.repository}'  | base64 -d > ./repository.tar.gz
         TUF_MIRROR=$(kubectl -n tuf-system get ksvc tuf -ojsonpath='{.status.url}')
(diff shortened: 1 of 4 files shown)


.github/workflows/fulcio-rekor-kind.yaml
```diff
@@ -203,6 +205,30 @@ jobs:
       run: |
         cosign verify --rekor-url ${{ env.REKOR_URL }} --allow-insecure-registry ${{ env.demoimage }}
 
+    # Test with cosign in 'airgapped mode'
+    # Uncomment these once modified cosign goes in.
```
k4leung4, line 209: do we have a commit or version we should look for?

k4leung4 (approved):

.github/workflows/fulcio-rekor-kind.yaml
vaikas: I'll grab the version after it gets committed. I tested with my own fork for functionality, but will need to update the instructions, documentation, so will do after it's been released. Is that ok?

vaikas: Sorry, I'll grab the 'release' once it gets released with my commit there.

PR #390, opened 2022-09-28
Add prober for verifying Rekor write correctness

### LINKED ISSUE #351: Add retrieval and verification of written entries to sigstore-prober
**Description**

When write probing is enabled, we can use the prober to verify consistency and correctness of newly written entries. Alternatively, we could modify the prober to run in "consistency" and "correctness" modes that verify different properties and emit metrics to track.

codysoyland:

#### Summary

This adds a verification flow to the prober to ensure new entries have valid verification material (inclusion proof and SET).

Fixes #351

### DIFF
--- a/cmd/prober/write.go
+++ b/cmd/prober/write.go
@@ -25,16 +25,17 @@ import (
 	"encoding/hex"
 	"encoding/json"
 	"fmt"
-	"io"
 	"net/http"
 	"time"
 
 	"github.com/go-openapi/strfmt"
 	"github.com/go-openapi/swag"
+	"github.com/prometheus/client_golang/prometheus"
 
 	"github.com/sigstore/cosign/pkg/cosign"
 	"github.com/sigstore/cosign/pkg/providers"
 	"github.com/sigstore/fulcio/pkg/api"
+	rclient "github.com/sigstore/rekor/pkg/client"
 	"github.com/sigstore/rekor/pkg/generated/models"
 	hashedrekord "github.com/sigstore/rekor/pkg/types/hashedrekord/v0.0.1"
 	"github.com/sigstore/sigstore/pkg/cryptoutils"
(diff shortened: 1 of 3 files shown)


cmd/prober/verification.go
```diff
@@ -0,0 +1,60 @@
...
+	"fmt"
+
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/sigstore/cosign/pkg/cosign"
+	rclient "github.com/sigstore/rekor/pkg/client"
+	"github.com/sigstore/rekor/pkg/generated/client/entries"
+	"github.com/sigstore/rekor/pkg/generated/models"
+)
+
+func rekorVerification(ctx context.Context) error {
```
priyawadhwa, line 28: Insetad of creating another entry, could we just do the verification on the entry we get in `rekorWriteEndpoint` in write.go?

cmd/prober/write.go
```diff
@@ -121,17 +122,33 @@ func rekorWriteEndpoint(ctx context.Context) error {
...
+	var logEntry models.LogEntry
+	err = json.NewDecoder(resp.Body).Decode(&logEntry)
+	if err != nil {
+		return fmt.Errorf("unmarshal: %w", err)
+	}
+	var logEntryAnon models.LogEntryAnon
+	for _, e := range logEntry {
+		logEntryAnon = e
+		break
+	}
```
codysoyland, line 145: Curious if there's a more elegant way to extract the `models.LogEntryAnon`... please let me know!

cmd/prober/write.go
```diff
@@ -121,17 +122,33 @@ func rekorWriteEndpoint(ctx context.Context) error {
...
 	defer resp.Body.Close()
-	body, _ = io.ReadAll(resp.Body)
-	fmt.Println(string(body))
-	return nil
+	var logEntry models.LogEntry
+	err = json.NewDecoder(resp.Body).Decode(&logEntry)
+	if err != nil {
+		return fmt.Errorf("unmarshal: %w", err)
+	}
```
codysoyland, line 140: Should we also test that we can re-fetch this entry from the log instead of just using the response from earlier? If so, should that be a new metric?

cmd/prober/write.go
priyawadhwa: this is probably the best way 😅

cmd/prober/write.go
priyawadhwa: we already test that we can fetch an entry from the log, so we should be good on that. with the current setup we can also be confident that if the verification succeeds that the entry exists, so we should be good!

priyawadhwa (approved):
nice!

codysoyland:
@priyawadhwa Thanks for the review! FYI, I don't have permission to merge.

PR #393, opened 2022-09-30
Add availability. SLO definitions to terraform for Fulcio and Rekor

var-sdk:
When enabled, this terraform creates custom monitoring services for Fulcio and Rekor and adds availability SLOs by server and prober metrics. We target 99.5% availability for supported methods and 80% availability for unsupported (redis index.) Alerts will be wired up in later changes.

SLO creation is off by default and must be enabled by setting the "create_slos" variable to true in the root sigstore module.


var-sdk:
cc @priyawadhwa

priyawadhwa (approved):

PR #395, opened 2022-09-30
Add prober flags for passing extra requests. 

var-sdk:

### DIFF
--- a/cmd/prober/endpoints.go
+++ b/cmd/prober/endpoints.go
@@ -20,47 +20,47 @@ var (
 )
 
 type ReadProberCheck struct {
-	endpoint string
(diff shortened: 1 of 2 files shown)


var-sdk:
@priyawadhwa wdyt about this approach for passing requests by flag? Would allow us to probe requests in staging that aren't yet ready in prod.

Based off the change in https://github.com/sigstore/scaffolding/pull/394 look at the second commit for the interesting change.

priyawadhwa (approved):
sgtm! also sorry i merged the other pr before looking at this one, i think you'll have to rebase.

codysoyland (approved):

PR #402, opened 2022-10-04
Create alert for prober verification failures

codysoyland:
Any single prober verification failure should yield an alert, as described in the ticket.

Related ticket: https://github.com/sigstore/scaffolding/issues/351
public-good-instance ticket: https://github.com/sigstore/public-good-instance/issues/711
Prober feature: https://github.com/sigstore/scaffolding/pull/390

### DIFF
--- a/terraform/gcp/modules/monitoring/prober/prober_alerts.tf
+++ b/terraform/gcp/modules/monitoring/prober/prober_alerts.tf
@@ -184,3 +184,37 @@ resource "google_monitoring_alert_policy" "prober_error_codes" {
   notification_channels = local.notification_channels
   project               = var.project_id
 }
+
+resource "google_monitoring_alert_policy" "prober_verification" {
+  alert_strategy {
+    auto_close = "604800s"
+  }
+
+  combiner = "OR"
+
+  conditions {
+    condition_threshold {
+      aggregations {
+        alignment_period   = "60s"
+        per_series_aligner = "ALIGN_SUM"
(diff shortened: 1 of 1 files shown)


terraform/gcp/modules/monitoring/prober/prober_alerts.tf
```diff
@@ -184,3 +184,37 @@ resource "google_monitoring_alert_policy" "prober_error_codes" {
...
+      trigger {
+        count   = "1"
+        percent = "0"
+      }
+    }
+
+    display_name = "Kubernetes Container - external/prometheus/verification"
+  }
+
+  display_name          = "API Prober: Verification failed at least once in the last 60s"
```
tstromberg, line 216: Normally I would associate a prober failure to be relating to reliability rather than correctness. 

Could the display name be adjusted to something that is more obvious to oncall responders? I may not have the terminology right, but I was thinking something like:

`Rekor write correctness: verifier returned 'false' within the last 60s`

tstromberg (changes requested):

terraform/gcp/modules/monitoring/prober/prober_alerts.tf
codysoyland: Good point! I took your proposal and prefixed with `API Prober` (for consistency) and added a documentation section with a rewording.

tstromberg (approved):

PR #405, opened 2022-10-06
Define basic ctlog and dex availability SLOs.

var-sdk:
For GA we will define basic server availability SLOs for all ctlog and dex services, as well as ctlog.AddPreChain, which is called during Fulcio certificate issuance. We can add prober SLOs if/when those endpoints are probed.


cpanato (approved):
cool

PR #406, opened 2022-10-06
Add allow lists for probed methods.

var-sdk:
Methods should be added to the allow list as they are promoted to supported rather than excluded by disallow list if accidentally probed.

### DIFF
--- a/terraform/gcp/modules/monitoring/prober/prober_alerts.tf
+++ b/terraform/gcp/modules/monitoring/prober/prober_alerts.tf
@@ -32,7 +32,7 @@ resource "google_monitoring_alert_policy" "prober_rekor_endpoint_latency" {
 
       comparison      = "COMPARISON_GT"
       duration        = "300s"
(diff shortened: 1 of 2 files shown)


terraform/gcp/modules/monitoring/prober/prober_alerts.tf
```diff
@@ -32,7 +32,7 @@ resource "google_monitoring_alert_policy" "prober_rekor_endpoint_latency" {
 
       comparison      = "COMPARISON_GT"
       duration        = "300s"
-      filter          = format("resource.type = \"prometheus_target\" AND metric.type = \"prometheus.googleapis.com/api_endpoint_latency/summary\" AND metric.labels.host = \"%s\"", var.rekor_url)
+      filter          = format("resource.type = \"prometheus_target\" AND metric.type = \"prometheus.googleapis.com/api_endpoint_latency/summary\" AND metric.labels.host = \"%s\" %s", var.rekor_url, local.rekor_endpoint_filter)
```
priyawadhwa, line 35: Should

```
AND metric.labels.host = \"%s\" %s"
```

be

```
AND metric.labels.host = \"%s\" AND %s"
```

var-sdk: Good catch!

vaikas (approved):

PR #416, opened 2022-10-10
Add explicit dependency on project_roles for all modules.

var-sdk:
This will allow us to set up necessary IAM grants to run the rest of the terraform as the first step in terraform.


cpanato (approved):
thank you

k4leung4 (approved):

PR #429, opened 2022-10-18
add support for pre-existing secrets for CTLog.

vaikas:

#### Summary
Because we've added keys to TUF already, we can't create them and then add them, we need to take the pre-existing
private key and then reuse that for creating CTLog config.

#### Release Note
* Add support for createctconfig to read a pre-existing private key to use for creating CTLog configuration.


haydentherapper (commented):
LGTM, I'll let someone with more knowledge of scaffolding approve

priyawadhwa (approved):

PR #435, opened 2022-10-20
drop k8s 1.22.x and add 1.25.x

k4leung4:

#### Summary

K8s 1.22.x is EOL next friday.
Drop testing for it, and add 1.25.x

### DIFF
--- a/hack/setup-kind.sh
+++ b/hack/setup-kind.sh
@@ -321,17 +321,6 @@ for x in $(kubectl get deploy --namespace knative-serving -oname); do
   kubectl rollout status --timeout 5m --namespace knative-serving "$x"
 done
 
-# Enable the features we need that are currently feature-flagged in Knative.
-# We do this last to ensure the webhook is up.
-while ! kubectl patch configmap/config-features \
-  --namespace knative-serving \
-  --type merge \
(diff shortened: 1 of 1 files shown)


cpanato:
if approved by the maintainers i will follow up to update the required checks

cpanato (dismissed):
lgtm

vaikas (dismissed):

vaikas:
FYI, the test:
```
Test github action with TUF / Test github action with TUF (v1.25.x, v0.4.6, 1.19, test github action
```

Failing is because 0.4.6 didn't have support yet for 1.25. Since we're now up to 0.4.12 in Scaffolding, I think there's more things that we could clean up from here since everything has been moved to using TUF based setup instead of the env variables.

.github/workflows/test-action-tuf.yaml
```diff
@@ -21,9 +21,9 @@ jobs:
       fail-fast: false # Keep running if one leg fails.
       matrix:
         k8s-version:
-        - v1.22.x
         - v1.23.x
         - v1.24.x
+        - v1.25.x
```
vaikas, line 26: This doesn't work with the v0.4.6 version below because that didn't support 1.25. I'd reckon drop the 1.25 test here for now, or change the v0.4.6 to test with something later (which will require changes in the community). Maybe easier to just drop this test for now, I can add it back later and clean up which versions we test with?

PR #437, opened 2022-10-22
plumb ways to add metadata about targets.

vaikas:

#### Summary
Start adding usage custom metadata for the TUF targets. Using same logic as here:
Inspired by this gist:
https://gist.github.com/haydentherapper/fd5fc3749f8f62e13de3ed08e8c671f3#file-byo-tuf-go-L109

### DIFF
--- a/pkg/repo/repo.go
+++ b/pkg/repo/repo.go
@@ -18,6 +18,7 @@ import (
 	"archive/tar"
 	"compress/gzip"
 	"context"
+	"encoding/json"
 	"fmt"
 	"io"
 	"io/fs"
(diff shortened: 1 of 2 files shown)


vaikas:
@haydentherapper would you mind taking a look.

cpanato (approved):
Lgtm

pkg/repo/repo.go
```diff
@@ -31,13 +32,26 @@ import (
...
+type TargetWithMetadata struct {
+	Name           string
+	Bytes          []byte
+	CustomMetadata []byte
+}
+
+type CustomMetadata struct {
+	Usage  string `json:"usage"`
+	Status string `json:"status"`
+}
```
haydentherapper, line 46: Since creating that script, `uri` was also added as custom metadata (for delegations I believe). See https://github.com/sigstore/root-signing/blob/7187cb661647a8420c1d2ebb354bc2b7b8cade37/ceremony/2022-10-18/repository/targets.json#L28 for an example

Can we add that here?

pkg/repo/repo.go
```diff
@@ -97,6 +108,43 @@ func CreateRepo(ctx context.Context, files map[string][]byte) (tuf.LocalStore, s
...
+func CreateRepo(ctx context.Context, files map[string][]byte) (tuf.LocalStore, string, error) {
+	targets := make([]TargetWithMetadata, 0, len(files))
+	for name, bytes := range files {
+		usage := ""
+		if strings.Contains(name, "fulcio") {
+			usage = "Fulcio"
+		} else if strings.Contains(name, "ctfe") {
+			usage = "CTFE"
+		} else {
+			usage = "Rekor"
```
haydentherapper, line 132: I'd change this to 

```
} else if strings.Contains(name, "rekor")
     usage = "Rekor"
} else {
     usage = "Unknown"
}
```

To match https://github.com/sigstore/root-signing/blob/7187cb661647a8420c1d2ebb354bc2b7b8cade37/ceremony/2022-10-18/repository/targets.json#L17

pkg/repo/repo.go
vaikas: yep, done

haydentherapper (approved):

PR #449, opened 2022-10-28
Terraform configuration for Timestamp Authority

haydentherapper:
Copy-paste for the win

Ref: https://github.com/sigstore/timestamp-authority/issues/48

Signed-off-by: Hayden Blauzvern <hblauzvern@google.com>

### DIFF
--- a/terraform/gcp/modules/sigstore/sigstore.tf
+++ b/terraform/gcp/modules/sigstore/sigstore.tf
@@ -257,6 +257,29 @@ module "fulcio" {
   ]
 }
 
+module "timestamp" {
+  source = "../timestamp"
+
+  region       = var.region
+  project_id   = var.project_id
+  cluster_name = var.cluster_name
+
+  // KMS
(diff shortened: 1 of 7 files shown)


haydentherapper:
One thing I wasn't sure about, do I need to configure anything for setting up K8s infra, or is that automatic due to depending on `module.gke-cluster`?

haydentherapper:
Also, how do I test this? In staging, or is there another option?

priyawadhwa:
> One thing I wasn't sure about, do I need to configure anything for setting up K8s infra

Nope, should be automatic with the cluster dependency!

I think testing in staging would probably be the quickest way to test this out, but it might be easier to test if you break up the PR a bit and incrementally apply the terraform changes (maybe timestamp module, then alerts/prober, then slos, then DNS?)

haydentherapper:
Updated to remove monitoring

priyawadhwa (approved):
looks great!

PR #463, opened 2022-11-08
add fulcio-grpc as knative service

bobcallaway:
Signed-off-by: Bob Callaway <bcallaway@google.com>

### DIFF
--- a/config/fulcio/fulcio/300-fulcio.yaml
+++ b/config/fulcio/fulcio/300-fulcio.yaml
@@ -20,13 +20,83 @@ spec:
       # This doesn't actually use Kubernetes credentials, so don't mount them in.
       automountServiceAccountToken: false
       containers:
-      - image: gcr.io/projectsigstore/fulcio@sha256:61081295a8f75ed7537b5d1f8c7320e078dc00e4562c0bf605fbefa062c690de # v0.5.3
+      - image: gcr.io/projectsigstore/fulcio@sha256:27c6e4fe64a72a537c133452d9c8e0518944d1d69aeee5e7ef8a9fbe70b8b5d3 # v1.0.0
         name: fulcio
         ports:
         - containerPort: 5555
         args:
           - "serve"
           - "--port=5555"
+          - "--ca=fileca"
+          - "--fileca-key"
+          - "/var/run/fulcio-secrets/key.pem"
+          - "--fileca-cert"
+          - "/var/run/fulcio-secrets/cert.pem"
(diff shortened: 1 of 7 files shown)


bobcallaway:
@vaikas I think once this is merged (assuming you're OK with it), we'll need to cut 0.4.13 release and then bump the references inline up to 0.4.13 and then cut another release or update the tag?

vaikas:
Thanks @bobcallaway I have no problem at all running parallel Fulcio, though shortly I'd like to still investigate the use the same port for both grpc/http :)

Something like this (that just doesn't rely on HTTP/2 for muxing):
https://github.com/chainguard-dev/go-grpc-kit

config/fulcio/fulcio/300-fulcio.yaml
```diff
@@ -20,7 +20,7 @@ spec:
       # This doesn't actually use Kubernetes credentials, so don't mount them in.
       automountServiceAccountToken: false
       containers:
-      - image: gcr.io/projectsigstore/fulcio@sha256:61081295a8f75ed7537b5d1f8c7320e078dc00e4562c0bf605fbefa062c690de # v0.5.3
+      - image: gcr.io/projectsigstore/fulcio@sha256:27c6e4fe64a72a537c133452d9c8e0518944d1d69aeee5e7ef8a9fbe70b8b5d3 # v1.0.0
```
vaikas, line 23: Thank you :)

vaikas (approved):
Thanks! What's the best way for us to test this?

config/fulcio/fulcio/300-fulcio.yaml
```diff
@@ -73,3 +73,73 @@ spec:
...
+  template:
+    metadata:
+      annotations:
+        autoscaling.knative.dev/min-scale: "1"
+    spec:
+      serviceAccountName: fulcio
+      # This doesn't actually use Kubernetes credentials, so don't mount them in.
+      automountServiceAccountToken: false
+      containers:
+      - image: gcr.io/projectsigstore/fulcio@sha256:27c6e4fe64a72a537c133452d9c8e0518944d1d69aeee5e7ef8a9fbe70b8b5d3 # v1.0.0
```
cpanato, line 93: why not use `gcr.io/projectsigstore/fulcio:v1.0.0@sha256:27c6e4fe64a72a537c133452d9c8e0518944d1d69aeee5e7ef8a9fbe70b8b5d3` and drop the comment

PR #482, opened 2022-11-23
Install TSA if release > 0.5.0. Remove old cruft.

vaikas:

#### Summary
Remove the legacy variables that have been gone for many versions now, only support TUF.
Remove old cruft
Add support for installing TSA for releases >= 0.5.0 (which will be cut after this merges).

#### Release Note
* Drop support for legacy env variables, only support TUF
* Add support for installing TSA for releases > v0.5.0

### DIFF
--- a/.github/workflows/test-release.yaml
+++ b/.github/workflows/test-release.yaml
@@ -30,7 +30,7 @@ jobs:
         - 1.19
 
     env:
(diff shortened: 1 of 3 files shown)


k4leung4 (approved):
Lgtm

PR #484, opened 2022-11-23
Exercise TSA in action/release tests.

vaikas:
#### Summary
Fulcio grpc was added _after_ 0.4.12, not in it, so fix it.
Exercise TSA in action / release tests.

### DIFF
--- a/.github/workflows/test-action-tuf.yaml
+++ b/.github/workflows/test-action-tuf.yaml
@@ -42,7 +42,6 @@ jobs:
       with:
         k8s-version: ${{ matrix.k8s-version }}
         version: ${{ matrix.release-version }}
-        legacy-variables: "false"
 
     # Install cosign
     - name: Install cosign
@@ -101,7 +100,7 @@ jobs:
         echo "myblob" > myblob
         if ! ./timestamp-cli --timestamp_server ${{ env.TSA_URL }} timestamp --hash sha256 --artifact myblob --out response.tsr ; then
           echo "failed to timestamp artifact"
-          exit =1
+          exit -1
         fi
         if ! ./timestamp-cli verify --timestamp response.tsr --artifact "myblob" --cert-chain ts_chain.pem ; then
           echo "failed to verify timestamp"
--- a/.github/workflows/test-action-tuf.yaml
+++ b/.github/workflows/test-action-tuf.yaml
@@ -85,6 +85,33 @@ jobs:
       run: |
         cosign verify --rekor-url ${{ env.REKOR_URL }} --allow-insecure-registry ${{ env.demoimage }}
 
+    - name: Checkout TSA for testing.
+      uses: actions/checkout@2541b1294d2704b0964813337f33b291d3f8596b # v3.0.2
+      with:
+        repository: sigstore/timestamp-authority
+        path: ./src/github.com/sigstore/timestamp-authority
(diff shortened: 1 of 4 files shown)


.github/workflows/test-release.yaml
```diff
@@ -131,6 +133,33 @@ jobs:
...
+      run: |
+        go build -o ./timestamp-cli ./cmd/timestamp-cli
+    - name: Exercise TSA
+      working-directory: ./src/github.com/sigstore/timestamp-authority
+      run: |
+        curl ${{ env.TSA_URL }}/api/v1/timestamp/certchain > ts_chain.pem
+        echo "myblob" > myblob
+        if ! ./timestamp-cli --timestamp_server ${{ env.TSA_URL }} timestamp --hash sha256 --artifact myblob --out response.tsr ; then
+          echo "failed to timestamp artifact"
+          exit =1
```
k4leung4, line 152: -1

k4leung4 (commented):
please also remove `legacy-variables: "false"` in `test-action-tuf.yaml`

.github/workflows/test-action-tuf.yaml
```diff
@@ -85,6 +85,33 @@ jobs:
...
+      run: |
+        go build -o ./timestamp-cli ./cmd/timestamp-cli
+    - name: Exercise TSA
+      working-directory: ./src/github.com/sigstore/timestamp-authority
+      run: |
+        curl ${{ env.TSA_URL }}/api/v1/timestamp/certchain > ts_chain.pem
+        echo "myblob" > myblob
+        if ! ./timestamp-cli --timestamp_server ${{ env.TSA_URL }} timestamp --hash sha256 --artifact myblob --out response.tsr ; then
+          echo "failed to timestamp artifact"
+          exit =1
```
k4leung4, line 104: -1

.github/workflows/test-action-tuf.yaml
```diff
@@ -85,6 +85,33 @@ jobs:
...
+        repository: sigstore/timestamp-authority
+        path: ./src/github.com/sigstore/timestamp-authority
+    - name: Build timestamp-cli
+      working-directory: ./src/github.com/sigstore/timestamp-authority
+      run: |
+        go build -o ./timestamp-cli ./cmd/timestamp-cli
+    - name: Exercise TSA
+      working-directory: ./src/github.com/sigstore/timestamp-authority
+      run: |
+        curl ${{ env.TSA_URL }}/api/v1/timestamp/certchain > ts_chain.pem
```
k4leung4, line 100: nit: should this be in a shell script file, so it can be called by `test-action-tuf.yaml` and `test-release.yaml`, without duplicating it?

vaikas: That's a great idea!

vaikas: hm. The only problem is that we don't have the artifact in some of these tests (like the action) since we don't actually grab the repo, and the same for testing the release.
So we would have to either:
 * check out the repo to get the bash script
 * make the bash script to a release artifact that we can pull
 
 WDYT about doing this separately?

k4leung4: separately sgtm

k4leung4 (approved):

PR #514, opened 2022-12-21
add module standalone_sqls for creating one off databases.

vaikas:

#### Summary
We need to create a standalone database so that we can migrate off of sql 8.0 to 5.7. So, in case we need to do this again, for whatever reason, created an array that behaves very similary to ctlog shard database instance creation.

#### Release Note
* Add module standalone_mysqls that allows for creating one off databases with `standalone` in the name.


haydentherapper (approved):

vaikas:
Knative failed to set up:
```
configmap/config-features patched
configmap/config-defaults patched (no change)
configmap/config-autoscaler patched (no change)
jq: error (at <stdin>:1): Cannot index string with string "ref"
error: no objects passed to apply
Error: Process completed with exit code 1.

```

k4leung4 (approved):

PR #516, opened 2022-12-30
Allow user of the action to explicitly specify INSTALL_TSA

k4leung4:
#### Summary

Allow users of the scaffolding setup action to explicitly specify INSTALL_TSA to allow disabling the installation of TSA with v0.5.0+

### DIFF
--- a/actions/setup/action.yml
+++ b/actions/setup/action.yml
@@ -50,6 +50,10 @@ inputs:
     description: 'kubernetes version to install (v1.23.x, v1.24.x, v1.25.x), default: v1.24.x'
     required: true
     default: 'v1.24.x'
+  install-tsa:
+    description: 'If not set, will auto-detect whether to install TSA, or "true" or "false" to explictly force install or not of TSA.'
+    required: false
+    default: ''
 runs:
   using: "composite"
   steps:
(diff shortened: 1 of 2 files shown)


priyawadhwa (approved):

cpanato (approved):
cool

actions/setup/action.yml
```diff
@@ -72,9 +72,11 @@ runs:
       # greater than v0.5.0 and install it.
       # the install process, so check to see if we are running >=5
       MINOR=$(echo $tag | cut -d '.' -f 2)
-      INSTALL_TSA="false"
-      if [ ${MINOR} -ge 5 ]; then
-        INSTALL_TSA="true"
+      if [[ -z "${INSTALL_TSA}" ]]; then
```
vaikas, line 75: I wonder if this would be better aligned as an input rather than an env variable?

vaikas:
@k4leung4 just curious why this was closed and not merged?

k4leung4:
sorry for not commenting the reason for closure.
i thought i needed it, but apparently the installing of tsa wasnt the issue i was running into.

if you think it is still useful, i can merge this, but i no longer have a use case for it.

actions/setup/action.yml
k4leung4: i thought about that, but then i still need to pass the env var down to the release setup script, so i thought it would be easier to just use an env var

vaikas:
ah, no worries was just curious, makes sense. Thanks!

actions/setup/action.yml
vaikas: Yeah, makes sense :)

PR #520, opened 2023-01-13
adding cloudsqlproxy wrapper and use exitdir to signal when the pod should terminate

cpanato:
#### Summary
- adding cloudsqlproxy wrapper and use exitdir to signal when the pod should terminate

in our public instance we use cloudsql and the cloudsql-proxy to connect in the DB
however in the `createdb` job the cloudsql run forever and never finish because it is not aware of the main container that creates the db, using the `https://github.com/chainguard-dev/exitdir` we can signal from `createdb` to cloudproxy when we are ready and then the job the gracefully finish

This is the first part of the fix and the second is to update https://github.com/sigstore/helm-charts/blob/main/charts/trillian/templates/createdb/createdb-job.yaml to use this image

This will fix: https://github.com/sigstore/public-good-instance/issues/965

### DIFF
--- a/.ko.yaml
+++ b/.ko.yaml
@@ -1,5 +1,7 @@
 ---
 defaultBaseImage: cgr.dev/chainguard/alpine-base:latest
+baseImageOverrides:
+  github.com/sigstore/scaffolding/cmd/cloudsqlproxy: gcr.io/cloudsql-docker/gce-proxy:1.32.0-alpine
 
 builds:
   - id: ctlog-createctconfig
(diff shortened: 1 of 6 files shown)


cpanato:
@vaikas @priyawadhwa let me know if this makes sense as initial iteration

vaikas (dismissed):
Super! Thanks for doing this!

bobcallaway (approved):

k4leung4 (approved):
sweet

.ko.yaml
```diff
@@ -1,182 +1,200 @@
+---
 defaultBaseImage: cgr.dev/chainguard/alpine-base:latest
+baseImageOverrides:
+  github.com/sigstore/scaffolding/cmd/cloudsqlproxy: gcr.io/cloudsql-docker/gce-proxy:1.32.0-alpine
```
k4leung4, line 4: should we use 1.33.1 here, as that is the default in the helm chart

https://github.com/sigstore/helm-charts/blob/fdca671033e29dc58e15234644cf511d146f72db/charts/trillian/Chart.yaml#LL37C14-L37C118

cpanato: yes, thanks for the catch!

cpanato: done, PTAL

vaikas (approved):

cpanato:
will continue the work

PR #539, opened 2023-01-30
Fix issue #1201, dedupe fulcio certs.

vaikas:

#### Summary
Fix configuration by deduping any possible entries in the config.

Fix: #1201


vaikas:
Wierd. Seems like rekor jobs are not running at all (or well, there are no pods for those).

haydentherapper (approved):
Do you have more background into how this occurred and was caught?

vaikas:
It's in a different repo :) I'll send you a link.

PR #542, opened 2023-02-07
Alerts for high kubernetes memory and cpu utilization

cdris:
Signed-off-by: Ceridwen Coghlan <cdriskill@google.com>

#### Summary

Adds alerting for utilization of requested memory and cpu > 90% per issue https://github.com/sigstore/public-good-instance/issues/1114

### DIFF
--- a/terraform/gcp/modules/monitoring/infra/alerts.tf
+++ b/terraform/gcp/modules/monitoring/infra/alerts.tf
@@ -246,3 +246,89 @@ resource "google_monitoring_alert_policy" "kms_crypto_request_alert" {
   notification_channels = local.notification_channels
   project               = var.project_id
 }
+
+### Kubernetes Alerts
+
+# Kubernetes Container Memory Utilization > 90%
+resource "google_monitoring_alert_policy" "k8s_container_memory_utilization" {
+  # In the absence of data, incident will auto-close in 7 days
+  alert_strategy {
+    auto_close = "604800s"
+  }
+
+  combiner = "OR"
+
+  conditions {
+    condition_threshold {
+      aggregations {
+        alignment_period   = "300s"
+        per_series_aligner = "ALIGN_MEAN"
+      }
+
+      comparison      = "COMPARISON_GT"
+      duration        = "0s"
+      filter          = "metric.type=\"kubernetes.io/container/memory/request_utilization\" resource.type=\"k8s_container\""
+      threshold_value = "0.9"
+
+      trigger {
+        count   = "1"
+        percent = "0"
+      }
+    }
+
+    display_name = "Kubernetes Container - Memory utilization [MEAN]"
+  }
+
+  display_name = "Kubernetes Container Memory Utilization > 90%"
+
+  documentation {
+    content   = "Kubernetes Container Memory Utilization is >90%. Please increase requested memory."
+    mime_type = "text/markdown"
+  }
+
+  enabled               = "true"
+  notification_channels = local.notification_channels
+  project               = var.project_id
+}
+
+# Kubernetes Container CPU Utilization > 90%
+resource "google_monitoring_alert_policy" "k8s_container_cpu_utilization" {
+  # In the absence of data, incident will auto-close in 7 days
+  alert_strategy {
+    auto_close = "604800s"
+  }
+
+  combiner = "OR"
+
+  conditions {
+    condition_threshold {
+      aggregations {
+        alignment_period   = "300s"
+        per_series_aligner = "ALIGN_MEAN"
+      }
+
+      comparison      = "COMPARISON_GT"
+      duration        = "0s"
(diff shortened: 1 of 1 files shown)


cdris:
@haydentherapper could you take a look and let me know your thoughts on using these metrics and whether they would have caught the prober oom? Wondering if I should stick with these container metrics or swap to using the allocatable utilization metrics on k8s nodes instead of the request utilization metrics for containers like I'm using here.

terraform/gcp/modules/monitoring/infra/alerts.tf
```diff
@@ -246,3 +246,89 @@ resource "google_monitoring_alert_policy" "kms_crypto_request_alert" {
...
+
+  conditions {
+    condition_threshold {
+      aggregations {
+        alignment_period   = "300s"
+        per_series_aligner = "ALIGN_MEAN"
+      }
+
+      comparison      = "COMPARISON_GT"
+      duration        = "0s"
```
haydentherapper, line 269: Do we want to increase the duration, say, to 5 minutes? Otherwise quick spikes will cause alerts that aren't actionable.

We should see what the previous duration was and confirm this alert would have fired during the previous outage.

terraform/gcp/modules/monitoring/infra/alerts.tf
```diff
@@ -246,3 +246,89 @@ resource "google_monitoring_alert_policy" "kms_crypto_request_alert" {
...
```
haydentherapper, line 311: same here

haydentherapper:
> could you take a look and let me know your thoughts on using these metrics and whether they would have caught the prober oom? Wondering if I should stick with these container metrics or swap to using the allocatable utilization metrics on k8s nodes instead of the request utilization metrics for containers like I'm using here.

These metrics should be good as long as we saw these metrics spike during the previous outage. Hopefully that data is still there, if so, can you confirm that this alert would have triggered? I think Cloud Monitoring will show you if your alerting threshold would have been met.

terraform/gcp/modules/monitoring/infra/alerts.tf
```diff
@@ -246,3 +246,89 @@ resource "google_monitoring_alert_policy" "kms_crypto_request_alert" {
...
+      filter          = "metric.type=\"kubernetes.io/container/memory/request_utilization\" resource.type=\"k8s_container\""
```
bobcallaway, line 270: I think this presumes we have requests set on all containers, otherwise this will never fire? 

Would we also want an alert if the actual memory of the nodes exceeds some thresholds (regardless of requested allocations)?

terraform/gcp/modules/monitoring/infra/alerts.tf
cdris: I based this on the cloud_sql_memory_utilization alert, this should do a mean aggregation over 5 minutes, and alert if that 5 minute mean is above the threshold. Or am I reading how alerting works incorrectly? Here's that cloudsql alert in gcp if it helps: https://console.cloud.google.com/monitoring/alerting/policies/3138531804338793217/edit?project=project-rekor

terraform/gcp/modules/monitoring/infra/alerts.tf
cdris: Updated to use the allocatable metric instead of request. We had also discussed potentially having a lower priority alert for limit utilization? I haven't been able to find any priority options for gcp alerts, do we currently have any low priority alerts? Do they just utilize different notification channels?

terraform/gcp/modules/monitoring/infra/alerts.tf
haydentherapper: You're correct, this looks good! Thanks for the link.

bobcallaway:
> These metrics should be good as long as we saw these metrics spike during the previous outage. Hopefully that data is still there, if so, can you confirm that this alert would have triggered? I think Cloud Monitoring will show you if your alerting threshold would have been met.

Were you able to do this @cdris? TF looks fine, but would be good to get confirmation that this would have caught the issue we were paged on.

cdris:
@bobcallaway our data retention is only 6 weeks as far as I can tell, so the relevant data was already out of retention before I started work on this. @haydentherapper unless you have more detailed notes on how high the cpu and mem usage got than what's referenced in [the incident](https://github.com/sigstore/public-good-instance/blob/main/incidents/2022-12-04-prober-oom.md)

haydentherapper:
IIRC CPU and mem usage were at 100%

Do you know if we can increase the data retention policy?

haydentherapper (approved):

cdris:
Per https://cloud.google.com/monitoring/quotas#data_retention_policy I'm not sure that retention period is modifiable?

cdris:
@bobcallaway just want to make sure I get your approval on this too before I merge

bobcallaway (approved):

PR #547, opened 2023-02-15
registry on port 5001 by default. From release uses 0.6.3.

vaikas:
#### Summary
Start registry by default on 5001 so it doesn't collide with airplay.
Update hack/setup-scaffolding-from-release to install 0.6.3 by default.
update readme

#### Release Note
Installed registry is on 5001 by default.


k4leung4 (approved):
thanks for the update and making it more user friendly

PR #565, opened 2023-03-06
Update deps

cpanato:
#### Summary
- Update deps
as soon cosign do a new release we use the release tag in the deps, for now using main head to unblock the upgrade


vaikas (approved):
Thank you!

PR #566, opened 2023-03-06
upgrade to use go1.20

cpanato:
#### Summary
- upgrade to use go1.20

### DIFF
--- a/.github/workflows/add-remove-new-fulcio.yaml
+++ b/.github/workflows/add-remove-new-fulcio.yaml
@@ -24,13 +24,13 @@ jobs:
       fail-fast: false # Keep running if one leg fails.
       matrix:
         k8s-version:
-        - v1.25.x
+          - v1.25.x
 
         leg:
-        - fulcio-key-rotation
+          - fulcio-key-rotation
 
         go-version:
-        - 1.18
+          - 1.20.x
 
     env:
       GOPATH: ${{ github.workspace }}
--- a/.github/workflows/test-action-tuf.yaml
+++ b/.github/workflows/test-action-tuf.yaml
@@ -21,15 +21,15 @@ jobs:
       fail-fast: false # Keep running if one leg fails.
       matrix:
         k8s-version:
-        - v1.23.x
-        - v1.24.x
-        - v1.25.x
+          - v1.23.x
+          - v1.24.x
+          - v1.25.x
         release-version:
-        - "latest-release" # Test explicitly with latest
+          - "latest-release" # Test explicitly with latest
         go-version:
-        - 1.19
+          - 1.20.x
         leg:
-        - test github action with TUF
+          - test github action with TUF
     env:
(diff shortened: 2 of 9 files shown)


.github/workflows/add-remove-new-fulcio.yaml
```diff
@@ -24,13 +24,15 @@ jobs:
...
 
         leg:
-        - fulcio-key-rotation
+          - fulcio-key-rotation
 
         go-version:
-        - 1.18
+          - 1.18
+          - 1.19
+          - '1.20'
```
k4leung4, line 35: @vaikas what is our golang version support policy

not sure if it is worthwhile to test all the go versions

k4leung4, line 35: nit: quote all the versions or none.

cpanato: it only require for `1.20` because yaml understands that as `1.2`

but will update all to use the ticks

k4leung4: if verify is only running with 1.20, i say we should only support the latest? and drop 1.18 and 1.19

cpanato: at least the latest then, i guess

vaikas: Yeah, good point @k4leung4
Since we do release artifacts and test with those, probably not that important to be testing with various versions of go.

cpanato:
will require a change to update the checks,doing that

cpanato:
ci checks changes: https://github.com/sigstore/community/pull/238

cpanato:
ci checks updated

vaikas (approved):

.github/workflows/test-action-tuf.yaml
```diff
@@ -21,15 +21,15 @@ jobs:
       fail-fast: false # Keep running if one leg fails.
       matrix:
         k8s-version:
-        - v1.23.x
-        - v1.24.x
-        - v1.25.x
+          - v1.23.x
```
k4leung4, line 24: should we drop 1.23? 
its eol on feb 28

vaikas: Yeah, well, this is a bit tricky because I know some folks have asked fairly recently for policy-controller support on something like 1.21, 1.22 so keeping testing as long as we can is something that I think would be nice. If there are breaking changes that would require making changes, then we can revisit them on case-by-case basis. Does that make sense? But I don't feel super strongly, just think it would be nice :)

vaikas: lol, it's a great question that we should have an answer to. Tricky bit is that indeed for us to run on older versions because we're relying on knative we will need to use the ENV variable to say it's ok to run. So 🤷

PR #567, opened 2023-03-08
expose variable to make timestamp module optional

bobcallaway:

### DIFF
--- a/terraform/gcp/modules/sigstore/sigstore.tf
+++ b/terraform/gcp/modules/sigstore/sigstore.tf
@@ -248,6 +248,10 @@ module "timestamp" {
   project_id   = var.project_id
   cluster_name = var.cluster_name
 
+  // Disable module entirely if timestamp
+  // is disabled
+  count = var.timestamp.enabled ? 1 : 0
+
   // KMS
   timestamp_keyring_name             = var.timestamp_keyring_name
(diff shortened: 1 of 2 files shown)


haydentherapper (approved):

k4leung4:
one side-effect of this change is that terraform resource path for these resources will change, and to avoid recreation (destroy and create), users will need to perform terraform move or terraform import to update the change in resource path

bobcallaway:
> one side-effect of this change is that terraform resource path for these resources will change, and to avoid recreation (destroy and create), users will need to perform terraform move or terraform import to update the change in resource path

OK - i'm open to other ways to implement this, but we will need to disable this as we don't want to run timestamp in staging/prod but I understand we want to leave this in for other deployment scenarios.

k4leung4:
> > one side-effect of this change is that terraform resource path for these resources will change, and to avoid recreation (destroy and create), users will need to perform terraform move or terraform import to update the change in resource path
> 
> OK - i'm open to other ways to implement this, but we will need to disable this as we don't want to run timestamp in staging/prod but I understand we want to leave this in for other deployment scenarios.

if it hasnt got into staging/prod yet, this is fine. or if we want this disabled for sigstore public instance, that is also fine.

just pointing out the caveat that this will be a breaking change for people that is using it already.

k4leung4 (approved):

Issue #622, opened 2023-05-03
Installing fulcio failed due to domain confusion


therealnb:
**Description**

Installing fulcio failed, the pods didn't come up. In the log

    2023-04-22T20:13:45.002Z	[31mFATAL[0m	app/serve.go:187	error loading --config-path=/etc/fulcio-config/config.json: provider https://kubernetes.default.svc: oidc: issuer did not match the issuer returned by provider, expected "https://kubernetes.default.svc" got "https://kubernetes.default.svc.cluster.local"

Had to change the config to add '.cluster.local'

    $ kubectl edit cm/fulcio-config

    apiVersion: v1
    data:
      config.json: |-
        {
          "OIDCIssuers": {
            "https://kubernetes.default.svc.cluster.local": {
              "IssuerURL": "https://kubernetes.default.svc.cluster.local",
              "ClientID": "sigstore",
              "Type": "kubernetes"
            }, ...

To get this to work.

**Version**
This was on the master branch sigs.k8s.io/release-utils/version.gitVersion=927786a-dirty

therealnb:
There is a PR to fix this here https://github.com/sigstore/scaffolding/pull/623.

PR #623, opened 2023-05-03
Change the domains in fulcio config map to match each other

therealnb:
#### Summary
This change is to address a startup issue in fulcio. See details at https://github.com/sigstore/scaffolding/issues/622

#### Release Note
This is a small change of config. This looks like it might have been fixed twice because there are sections in the scripts like this
```
# Install Fulcio and wait for it to come up
echo '::group:: Install Fulcio'
if [ "${NEED_TO_UPDATE_FULCIO_CONFIG}" == "true" ]; then
  echo "Fixing Fulcio config"
  cp config/fulcio/fulcio/200-configmap.yaml ./200-configmap.yaml
  # The sed works differently in mac and other places, so just shuffle
  # files around for now.
  sed 's@https://kubernetes.default.svc.cluster.local@https://kubernetes.default.svc@' config/fulcio/fulcio/200-configmap.yaml > ./200-configmap-new.yaml
  mv ./200-configmap-new.yaml config/fulcio/fulcio/200-configmap.yaml
fi
make ko-apply-fulcio
echo '::endgroup::'
```
Where the change in this PR would be reverted for k8s < 1.24.

CC @vaikas

### DIFF
--- a/config/fulcio/fulcio/200-configmap.yaml
+++ b/config/fulcio/fulcio/200-configmap.yaml
@@ -9,7 +9,7 @@ data:
     {
       "OIDCIssuers": {
         "https://kubernetes.default.svc.cluster.local": {
-          "IssuerURL": "https://kubernetes.default.svc",
+          "IssuerURL": "https://kubernetes.default.svc.cluster.local",
           "ClientID": "sigstore",
           "Type": "kubernetes"
         },


therealnb:
Hmmm, different versions different behaviour. Needs investigation...

haydentherapper:
This was completed at some point.

PR #652, opened 2023-05-30
Labeling severity for low priority alerts

cdris:
#### Summary

PagerDuty now understands manually labeled severities. By default, alerts will fire with critical severity and high urgency, but we can manually label alerts with severities of "warning" or "info" for alerts which should have low urgency for the purposes of paging oncallers.


cdris:
Will do

PR #657, opened 2023-06-08
Only creating trillian db if given instance name

cdris:
#### Summary

Per https://github.com/sigstore/public-good-instance/issues/1397 staging terraform state is currently broken because sigstore-staging-mysql-9a4f04b6 was deleted via the gcp console (https://github.com/sigstore/public-good-instance/issues/1190). We're going to manually delete this resource from the terraform state, but once we do, applying the current terraform will just generate a new trillian db instance with a new randomly generated suffix/name. This is no longer necessary because staging now has a new trillian db instance configured by different resources (see [here](https://github.com/sigstore/public-good-instance/issues/1397#issuecomment-1579401017) for all of the links). But! We cannot delete these resources because these are the same resource blocks defining the trillian instance in prod. I believe this mismatch in staging and prod came about due to https://github.com/sigstore/public-good-instance/issues/708.

So TL;DR here's my attempt at conditionally creating the trillian instance and dependent resources only if instance_name is not empty (i.e. prod). Happily accepting comments from anyone who knows terraform as I'm not confident in how this works with all of the subsequent dependent resources.

### DIFF
--- a/terraform/gcp/modules/mysql/mysql.tf
+++ b/terraform/gcp/modules/mysql/mysql.tf
@@ -78,7 +78,7 @@ resource "google_service_account_iam_member" "gke_sa_iam_member_trillian_logserv
 }
 
 resource "google_project_iam_member" "logserver_iam" {
-  # // Give trillian logserver permission to export metrics to Stackdriver
+  // Give trillian logserver permission to export metrics to Stackdriver
   for_each = toset([
     "roles/logging.logWriter",
     "roles/monitoring.metricWriter",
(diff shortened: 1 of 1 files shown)


cdris:
CC @bobcallaway @vaikas since you'll have more context on the deletion of the db and the separate terraform resources that create the current db in staging.

cdris:
Also if anyone has better thoughts on conditional resources in terraform, they'd be greatly appreciated.

cdris:
@vaikas what would happen if we made [this](https://github.com/sigstore/public-good-instance/blob/345d2707a6fe9bb0b21f3185cc9a46de654f035a/terraform/environments/staging/1-infrastructure/general.auto.tfvars#L66) an empty list and set mysql_instance_name to the current staging instance name like [prod](https://github.com/sigstore/public-good-instance/blob/6416019a9a8fcb52b8335aeca055ef38a43c9ecf/terraform/environments/production/1-infrastructure/production.tf#LL219C3-L219C22) does?

vaikas:
Sorry for the tardy response...

As to the very specific question about 'what would happen', I think we could use the terraform plan maybe to see what the actions would be.

The reason I didn't delete the database earlier was precisely because I was worried what it would do to the terraform. Maybe one way to try to make the smallest change would be for us to have terraform forget about the deleted database. I wonder if we could use some combination of these:

https://developer.hashicorp.com/terraform/cli/commands/state/rm

haydentherapper:
Agree about deleting the resource first, @pwelch was gonna try that. How about this:
*Delete resource, run plan to see what happens
*If plan will recreate the db, try @cdris’s proposal. I think that should work, since prod has the same set up.

cdris:
@pwelch is doing the delete now, zoom invite in slack if you'd like to join

pwelch:
Update, we didn't get to this on Friday because it requires higher permissions. I am not added to the correct role and looking into this today/this week.

cdris:
Closing this as https://github.com/sigstore/public-good-instance/issues/1397 has now been resolved

PR #701, opened 2023-07-16
increase versions for rekor, fulcio, trillian, tsa to match latest

bobcallaway:
We really need some automation or triggers to bump these more regularly


haydentherapper (approved):
Could you file an issue to track adding automation?

cpanato:
we maybe can use the same we are using in the public-instance repo, if you agree i can work on that

cpanato (approved):

PR #703, opened 2023-07-17
Fix repo tmp path for running as non-root user

mstrangfeld:
#### Summary

I had the problem that I need to run the TUF server as nonroot on Kubernetes. This tiny change fixes some permission issues.


cpanato (approved):

vaikas:
Thank you!

PR #704, opened 2023-07-21
change issuerURL for fulcio config map

bobcallaway:

### DIFF
--- a/config/fulcio/fulcio/200-configmap.yaml
+++ b/config/fulcio/fulcio/200-configmap.yaml
@@ -9,7 +9,7 @@ data:
     {
       "OIDCIssuers": {
         "https://kubernetes.default.svc.cluster.local": {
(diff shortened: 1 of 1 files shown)


priyawadhwa (approved):

k4leung4:
please rebase

vaikas:
FWIW I had run into this in the past:
https://github.com/sigstore/scaffolding/commit/2e198ea9a952e85701e7c962dc8bae4c0629e233#diff-3d56f080bbe91814f938f3fc46dac11063873b04c71d32adc9c3315360b40be4

And this is controlled for better or worse here:
https://github.com/sigstore/scaffolding/blob/main/hack/setup-scaffolding.sh#L20

So wondering if we may be able to add more conditionals here?
https://github.com/sigstore/scaffolding/blob/main/hack/setup-scaffolding.sh#L25

@bobcallaway

Issue #699, opened 2023-07-14
tests should be scoped to terraform or k8s or both


bobcallaway:
**Description**

Currently terraform-only PRs are blocked for 20+ minutes waiting on k8s-based tests that are not relevant to the changes at hand.

PR #705, opened 2023-07-21
Bump default knative version to 1.10, remove k8s 1.23

wlynch:
#### Summary

Version skew is high enough in Kubernetes that there are CRDs being removed that make the k8s + knative versions incompatible.

Standardize on supported k8s versions: https://kubernetes.io/releases/ (>= 1.24).

Default knative to v1.10.0 (which has a MIN_KUBERNETES_VERSION of 1.24): https://github.com/knative/community/blob/main/mechanics/RELEASE-SCHEDULE.md#upcoming-releases

Undo issuer change made in https://github.com/sigstore/scaffolding/commit/0bcfba714c4e78858b0828821042a0c96066c500

#### Release Note

Dropped support for Kubernetes 1.23

PR #715, opened 2023-07-25
filter tests to save time on terraform-only PRs

bobcallaway:
this change adds `paths-ignore` filters to make terraform-only changes run faster

Fixes: #699

### DIFF
--- a/.github/workflows/test-release.yaml
+++ b/.github/workflows/test-release.yaml
@@ -3,6 +3,8 @@ name: Fulcio&Rekor E2E Tests Using Release
 on:
   pull_request:
     branches: [ main ]
+    paths-ignore:
+      - 'terraform/**'
 
 defaults:
   run:
--- a/.github/workflows/add-remove-new-fulcio.yaml
+++ b/.github/workflows/add-remove-new-fulcio.yaml
@@ -3,6 +3,8 @@ name: Key rotation for Fulcio E2E Tests
 on:
   pull_request:
     branches: [ main ]
+    paths-ignore:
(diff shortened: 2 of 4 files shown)


haydentherapper (dismissed):

haydentherapper (dismissed):

.github/workflows/test-release.yaml
```diff
@@ -52,9 +55,10 @@ jobs:
 
     - name: Setup Cluster
       # TODO: update after next release.
-      uses: chainguard-dev/actions/setup-kind@52eca3baf7c09ec1dbc4195449779fa59cc618e6
+      uses: bobcallaway/cg-actions/setup-kind@f807492463bc85bc56ddbac9ddc265e3caa4c8ce
```
hectorj2f, line 58: why do we need to use this ?

hectorj2f:
@bobcallaway I am facing the same issue, happy to help fixing it! I was looking at https://github.com/sigstore/scaffolding/pull/704 or https://github.com/sigstore/scaffolding/pull/705 after talking with @wlynch.

vaikas:
Since this is fixing the bug of not running tests unnecessarlly, I'm wondering if we should consider:
 * Not running tests if Terraform files are only changed
 * Maybe running the release tests not on every PR but every now and then. When I added them, I just wanted to prevent breaking users that are using a particular release with unintentional changes. This was (probably?) more important when things in sigstore land were changing Fast&Furious (is it up to 9 🤔 ?)

hectorj2f (approved):

PR #720, opened 2023-08-07
Fix broken tuf presubmit

k4leung4:
#### Summary

try fixing broken tuf presubmit

presubmit fulcio failing with

> FATAL[0m	app/serve.go:209	error loading --config-path=/etc/fulcio-config/config.json: provider https://kubernetes.default.svc: oidc: issuer did not match the issuer returned by provider, expected "https://kubernetes.default.svc" got "https://kubernetes.default.svc.cluster.local"

The presubmit uses the latest release shell script, so added a temp hack to sed needed fix and checking in the fix to the shell script that would be picked up with the next release.

My guess is this is triggered by the last release picking up fulcio v1.3.4, was v1.0.0 before


cpanato (dismissed):

vaikas (approved):
Thank you!

PR #722, opened 2023-08-09
Specify the service account issuer for kubernetes clusters

hectorj2f:
#### Summary

I've made some changes to setup-kind actions to set a specific service account issuer. This might fix the current mismatch for the OIDC kubernetes provider when deploying the latest version on a kind cluster.

### DIFF
--- a/hack/setup-scaffolding-from-release.sh
+++ b/hack/setup-scaffolding-from-release.sh
@@ -95,7 +95,7 @@ if [ "${NEED_TO_UPDATE_FULCIO_CONFIG}" == "true" ]; then
   echo "Fixing Fulcio config for < 1.23.X Kubernetes"
   curl -Ls "${FULCIO}" | sed 's@https://kubernetes.default.svc.cluster.local@https://kubernetes.default.svc@' | kubectl apply -f -
 else
-  curl -Ls "${FULCIO}" | sed 's@"IssuerURL": "https://kubernetes.default.svc",@"IssuerURL": "https://kubernetes.default.svc.cluster.local",@' | kubectl apply -f -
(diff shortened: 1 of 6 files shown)


hack/setup-scaffolding-from-release.sh
```diff
@@ -95,7 +95,7 @@ if [ "${NEED_TO_UPDATE_FULCIO_CONFIG}" == "true" ]; then
   echo "Fixing Fulcio config for < 1.23.X Kubernetes"
   curl -Ls "${FULCIO}" | sed 's@https://kubernetes.default.svc.cluster.local@https://kubernetes.default.svc@' | kubectl apply -f -
 else
-  curl -Ls "${FULCIO}" | sed 's@"IssuerURL": "https://kubernetes.default.svc",@"IssuerURL": "https://kubernetes.default.svc.cluster.local",@' | kubectl apply -f -
+  curl -Ls "${FULCIO}" | sed 's@"IssuerURL": "https://kubernetes.default.svc.cluster.local",@"IssuerURL": "https://kubernetes.default.svc.cluster.local",@' | kubectl apply -f -
```
k4leung4, line 98: we should be able to replace this line, as your change would be replacing it with the same thing

k4leung4 (approved):

hack/setup-scaffolding-from-release.sh
hectorj2f: True! I missed this!

k4leung4 (approved):
thanks for fixing this

bobcallaway (approved):

wlynch (approved):

PR #729, opened 2023-08-19
expose GKE monitoring components as variable

bobcallaway:
exposes cluster config parameter for monitoring components per https://cloud.google.com/stackdriver/docs/solutions/gke/managing-metrics#control-plane-metrics

### DIFF
--- a/terraform/gcp/modules/gke_cluster/variables.tf
+++ b/terraform/gcp/modules/gke_cluster/variables.tf
@@ -216,6 +216,6 @@ variable "managed_prometheus" {
 }
 
 variable "monitoring_components" {
-  type    = list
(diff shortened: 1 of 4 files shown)


k4leung4 (approved):

terraform/gcp/modules/gke_cluster/variables.tf
```diff
@@ -214,3 +214,8 @@ variable "managed_prometheus" {
   type    = bool
   default = true
 }
+
+variable "monitoring_components" {
+  type    = list(any)
```
k4leung4, line 219: nit: you can also use `list(string)`

haydentherapper (approved):

PR #751, opened 2023-09-07
disable codeql on terraform-only changes

bobcallaway:

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -7,6 +7,8 @@ on:
   pull_request:
     branches:
       - main
+    paths-ignore:
+      - 'terraform/**'
   schedule:
     - cron: '32 8 * * 2'
 


cpanato (approved):

k4leung4 (approved):

.github/workflows/codeql-analysis.yml
```diff
@@ -7,6 +7,8 @@ on:
   pull_request:
     branches:
       - main
+    paths-ignore:
+      - 'terraform/**'
```
cpanato, line 11: humm that will not run and the check that is required will not be valid and will block

we should use the `dorny/paths-filter@v2` to filter then that will properly skip the job

cpanato (changes requested):
sorry double check that

.github/workflows/codeql-analysis.yml
bobcallaway: hmm, its worked across other workflows in this repo....... I guess I don't understand the difference?

cpanato: Oh really? That is good news then GitHub changed that behavior nice

cpanato (approved):

PR #757, opened 2023-09-14
Add alert for Cloud SQL CPU utilization > 90% for 300s

codysoyland:
Fixes https://github.com/sigstore/public-good-instance/issues/1649

PR #758, opened 2023-09-15
Fix CodeQL path filtering

codysoyland:
This reverts commit 4d6984b8dd70ba2abb0da91b70dc682f2ee310a4.

Branch protection rules require the CodeQL check. This change uses `dorny/paths-filter@v2` to filter the path as a step in the workflow so that the required check still runs.

This blocks https://github.com/sigstore/scaffolding/pull/757 which is a terraform-only change.

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -30,14 +30,27 @@ jobs:
       - name: Checkout repository
         uses: actions/checkout@3df4ab11eba7bda6032a0b82a6bb43b11571feac # v4.0.0
 
+      - name: Filter paths
+        uses: dorny/paths-filter@v2
+        id: changes
+        with:
+          filters: |
+            gocode:
+              - 'pkg/**'
+              - 'cmd/**'
+
       # Initializes the CodeQL tools for scanning.
       - name: Initialize CodeQL
+        if: steps.changes.outputs.gocode == 'true'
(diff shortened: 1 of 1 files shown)


cpanato:
we need to use this action to filter out https://github.com/sigstore/scaffolding/pull/751#discussion_r1318861448

i can submit a PR later

codysoyland:
> we need to use this action to filter out [#751 (comment)](https://github.com/sigstore/scaffolding/pull/751#discussion_r1318861448)
> 
> i can submit a PR later

Good call, I can do it in this PR actually.

cpanato (approved):
thank you!

codysoyland:
@cpanato Do you understand why the bare `CodeQL` status check on this PR is not running? I don't know where that came from, as the `CodeQL / Analyze (go)` check is running/passing.

cpanato:
> @cpanato Do you understand why the bare `CodeQL` status check on this PR is not running? I don't know where that came from, as the `CodeQL / Analyze (go)` check is running/passing.

we will need to update that, gimme a sec

cpanato:
lets do this: https://github.com/sigstore/community/pull/321

and later I will run a test pr to check others requires that we can add, lets apply that to unblock you

bobcallaway:
Thanks all, i appreciate you cleaning up my mistake

PR #801, opened 2023-10-25
set default autoscaling service_account to not be 'default'

bobcallaway:
@vaikas @yrobla I was trying to look at https://github.com/sigstore/public-good-instance/issues/1295 and https://github.com/sigstore/public-good-instance/issues/1301 - I think this will stop using the `default` SA for autoscaling and use the one we wanted instead. PTAL and let me know what you think?


vaikas (approved):
Awesomesauce! So we should see how it looks in the planning, so might as well get this in.

bobcallaway:
> Awesomesauce! So we should see how it looks in the planning, so might as well get this in.

it's worth a shot :)

PR #814, opened 2023-11-04
clean up terraform for post-install modules

bobcallaway:
A few changes:

- Remove providers that are not referenced in the module(s)
- Relax constraints to allow upgrades
- Upgrades `hashicorp/helm` provder to latest (`kubectl` provider is already at latest)
- Remove label from `trillian_namespace`, as the [helm-chart we use does not specify this](https://github.com/sigstore/helm-charts/blob/main/charts/trillian/templates/namespace.yaml ) and argocd reverts it immediately after terraform applies the change, and then we constantly get an "update in-place" message while running `terraform plan`

### DIFF
--- a/terraform/gcp/modules/argocd/versions.tf
+++ b/terraform/gcp/modules/argocd/versions.tf
@@ -15,28 +15,16 @@
  */
 
 terraform {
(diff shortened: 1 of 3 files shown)


cpanato (approved):
cool

PR #855, opened 2023-11-14
Add support for specifying a different mysql version for each ctlog shard

priyawadhwa:
we can use this when we spin up a new ct log shard with mysql 8 to test out trillian/mysql 8 in staging.

### DIFF
--- a/terraform/gcp/modules/sigstore/sigstore.tf
+++ b/terraform/gcp/modules/sigstore/sigstore.tf
@@ -329,7 +329,7 @@ module "ctlog" {
 module "ctlog_shards" {
   source = "../mysql-shard"
 
-  for_each = toset(var.ctlog_shards)
+  for_each = var.ctlog_shards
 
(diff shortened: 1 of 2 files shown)


terraform/gcp/modules/sigstore/sigstore.tf
```diff
@@ -339,7 +339,7 @@ module "ctlog_shards" {
   cluster_name = var.cluster_name
   // NB: These are commented out so that we pick up the defaults
```
haydentherapper, line 340: nit: Can you move the comment below this line?

haydentherapper (approved):
That was straightforward! Thanks!

terraform/gcp/modules/sigstore/sigstore.tf
priyawadhwa: done!

haydentherapper (approved):

PR #886, opened 2023-12-05
chore: improve error logs for prober

hectorj2f:
#### Summary

@bobcallaway found that the error logs of the prober weren't giving much information of what was happening. We also need to check the response code.

### DIFF
--- a/cmd/prober/write.go
+++ b/cmd/prober/write.go
@@ -163,14 +163,16 @@ func rekorWriteEndpoint(ctx context.Context, cert *x509.Certificate, priv *ecdsa
 	defer resp.Body.Close()
 	exportDataToPrometheus(resp, rekorURL, endpoint, POST, latency)
 
-	if resp.StatusCode != http.StatusOK {
-		return fmt.Errorf("invalid status code when checking an entry in rekor: %s", resp.Status)
+	if resp.StatusCode != http.StatusCreated {
+		body, _ := io.ReadAll(resp.Body)
+		return fmt.Errorf("invalid status code '%s' when checking an entry in rekor with body '%s'", resp.Status, string(body))
 	}
 	// If entry was added successfully, we should verify it
 	var logEntry models.LogEntry
 	err = json.NewDecoder(resp.Body).Decode(&logEntry)
 	if err != nil {
-		return fmt.Errorf("error decoding the log entry with body '%v' and error: %w", resp.Body, err)
+		body, _ := io.ReadAll(resp.Body)
+		return fmt.Errorf("error decoding the log entry with body '%s' and error: %w", string(body), err)
 	}
 	var logEntryAnon models.LogEntryAnon
 	for _, e := range logEntry {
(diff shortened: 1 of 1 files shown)


cmd/prober/write.go
```diff
@@ -163,11 +163,14 @@ func rekorWriteEndpoint(ctx context.Context, cert *x509.Certificate, priv *ecdsa
 	defer resp.Body.Close()
 	exportDataToPrometheus(resp, rekorURL, endpoint, POST, latency)
 
+	if resp.StatusCode != http.StatusOK {
```
bobcallaway, line 166: I think you will need `http.StatusCreated` since this is a write path.

cmd/prober/write.go
```diff
@@ -163,11 +163,14 @@ func rekorWriteEndpoint(ctx context.Context, cert *x509.Certificate, priv *ecdsa
...
+		return fmt.Errorf("invalid status code when checking an entry in rekor: %s", resp.Status)
```
bobcallaway, line 167: we should probably try to read all of `resp.Body` (a `io.Reader`) and include that in this error message

cmd/prober/write.go
```diff
@@ -163,11 +163,14 @@ func rekorWriteEndpoint(ctx context.Context, cert *x509.Certificate, priv *ecdsa
...
+	}
 	// If entry was added successfully, we should verify it
 	var logEntry models.LogEntry
 	err = json.NewDecoder(resp.Body).Decode(&logEntry)
 	if err != nil {
-		return fmt.Errorf("unmarshal: %w", err)
+		return fmt.Errorf("error decoding the log entry with body '%v' and error: %w", resp.Body, err)
```
bobcallaway, line 173: we should probably try to read all of `resp.Body` (a `io.Reader`) and include that in this error message

otherwise we'll just get a pointer printed here instead of the full body text

cmd/prober/write.go
```diff
@@ -163,11 +163,16 @@ func rekorWriteEndpoint(ctx context.Context, cert *x509.Certificate, priv *ecdsa
 	defer resp.Body.Close()
 	exportDataToPrometheus(resp, rekorURL, endpoint, POST, latency)
 
+	if resp.StatusCode != http.StatusCreated {
+		body, _ := io.ReadAll(resp.Body)
+		return fmt.Errorf("invalid status code '%s' when checking an entry in rekor with body '%s'", resp.Status, body)
```
bobcallaway, line 168:
```suggestion
		return fmt.Errorf("invalid status code '%s' when checking an entry in rekor with body '%s'", resp.Status, string(body))
```

cmd/prober/write.go
```diff
@@ -163,11 +163,16 @@ func rekorWriteEndpoint(ctx context.Context, cert *x509.Certificate, priv *ecdsa
...
+	}
 	// If entry was added successfully, we should verify it
 	var logEntry models.LogEntry
 	err = json.NewDecoder(resp.Body).Decode(&logEntry)
 	if err != nil {
-		return fmt.Errorf("unmarshal: %w", err)
+		body, _ := io.ReadAll(resp.Body)
+		return fmt.Errorf("error decoding the log entry with body '%s' and error: %w", body, err)
```
bobcallaway, line 175:
```suggestion
		return fmt.Errorf("error decoding the log entry with body '%s' and error: %w", string(body), err)
```

haydentherapper (approved):

PR #944, opened 2024-01-17
Create KMS and service account resources for TUF-on-CI

haydentherapper:
This adds a service account (which will be granted access to a workload identity pool in the private repo), a KMS keyring and key, and grants signerVerifier to the service account.

### DIFF
--- a/terraform/gcp/modules/tuf/kms.tf
+++ b/terraform/gcp/modules/tuf/kms.tf
@@ -1,0 +1,40 @@
+/**
+ * Copyright 2023 The Sigstore Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+resource "google_kms_key_ring" "tuf-keyring" {
+  name     = var.tuf_keyring_name
+  location = var.kms_location
+  project  = var.project_id
+}
+
+resource "google_kms_crypto_key" "tuf-key" {
+  name     = var.tuf_key_name
+  key_ring = google_kms_key_ring.tuf-keyring.id
+  purpose  = "ASYMMETRIC_SIGN"
+  version_template {
+    algorithm        = "EC_SIGN_P384_SHA384"
+    protection_level = "SOFTWARE"
+  }
+
+  depends_on = [google_kms_key_ring.tuf-keyring]
+}
+
+resource "google_kms_key_ring_iam_member" "tuf-sa-key-iam" {
+  key_ring_id = google_kms_key_ring.tuf-keyring.id
+  role        = "roles/cloudkms.signerVerifier"
+  member      = format("serviceAccount:%s@%s.iam.gserviceaccount.com", var.tuf_service_account_name, var.project_id)
+  depends_on  = [google_kms_key_ring.tuf-keyring, google_service_account.tuf-sa]
+}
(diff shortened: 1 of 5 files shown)


terraform/gcp/modules/tuf/kms.tf
```diff
@@ -0,0 +1,40 @@
+/**
+ * Copyright 2023 The Sigstore Authors
```
cpanato, line 2:
```suggestion
 * Copyright 2024 The Sigstore Authors
```

terraform/gcp/modules/tuf/service_accounts.tf
```diff
@@ -0,0 +1,21 @@
+/**
+ * Copyright 2023 The Sigstore Authors
```
cpanato, line 2:
```suggestion
 * Copyright 2024 The Sigstore Authors
```

cpanato (commented):
lgtm

terraform/gcp/modules/tuf/kms.tf
```diff
@@ -0,0 +1,40 @@
...
+
+  depends_on = [google_kms_key_ring.tuf-keyring]
+}
+
+resource "google_kms_key_ring_iam_member" "tuf-sa-key-iam" {
+  key_ring_id = google_kms_key_ring.tuf-keyring.id
+  role        = "roles/cloudkms.signerVerifier"
+  member      = format("serviceAccount:%s@%s.iam.gserviceaccount.com", var.tuf_service_account_name, var.project_id)
+  depends_on  = [google_kms_key_ring.tuf-keyring, google_service_account.tuf-sa]
+}
```
jku, line 40: this looks good to me. 

If you decide to remove the service account variable as suggested in other comment, member here would need changing to something like 
`member = "serviceAccount:${google_service_account.tuf-sa.email}"
`

jku (approved):
Looks good to me.  I'm ok with merging this but I have two potential suggestions for consideration:
* are the variables useful (What's the advantage of choosing non-default sa name or keyring name) ? I left some specific suggestions in code
* this one I can do in a future PR, no need to include in this one: It would be useful to add public key viewer permissions, something like
    ```
    resource "google_kms_key_ring_iam_member" "tuf-everyone-key-iam" {
      key_ring_id = google_kms_key_ring.tuf-keyring.id
      role        = "roles/cloudkms.publicKeyViewer"
      member      = "allUsers"
      depends_on  = [google_kms_key_ring.tuf-keyring]
    }
    ```
    (member could be limited to all project members if there's a group like that, but I think allUsers should be fine too...)

jku:
Some details for others following this:
* this configuration is added to enable https://github.com/sigstore/root-signing-staging right now
* the goal is to have contained configuration that manages the GCP resource that a "root-signing" installation (staging or not) requires 
* as this is now in a reusable module this config will be usable by (production) root-signing itself later on, but making that happen will require changes in root-signing: currently the equivalent parts in root-signing use several different service accounts and are configured partly manually because historical reasons

terraform/gcp/modules/tuf/service_accounts.tf
```diff
@@ -0,0 +1,21 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+resource "google_service_account" "tuf-sa" {
+  account_id   = var.tuf_service_account_name
```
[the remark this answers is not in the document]
haydentherapper: `cluster_name` is related to GKE. I'm not sure what we could parameterize on beyond project name. I decided to add variables for all new fields just in case someone was opinionated on names, but there's a default value for each so no configuration is needed.

haydentherapper:
> are the variables useful (What's the advantage of choosing non-default sa name or keyring name) ? I left some specific suggestions in code

Left comments in line, not a must to have, just was easy to add in case someone wants to customize names

> this one I can do in a future PR, no need to include in this PR: It would be useful to add public key viewer permissions, something like

For opening up viewer permissions on the key, I've added a variable to specify members who can view the key.

bobcallaway (approved):

PR #947, opened 2024-01-19
Grant TUF SA access to modify storage objects

haydentherapper:
Needed to use the TUF SA on GitHub Actions workflows.

### DIFF
--- a/terraform/gcp/modules/tuf/tuf.tf
+++ b/terraform/gcp/modules/tuf/tuf.tf
@@ -74,6 +74,14 @@ resource "google_storage_bucket_iam_member" "public_tuf_member" {
   depends_on = [google_storage_bucket.tuf]
 }
 
+resource "google_storage_bucket_iam_member" "tuf_sa_editor" {
+  bucket = google_storage_bucket.tuf.name
+  role   = "roles/storage.objectUser"
+  member = format("serviceAccount:%s@%s.iam.gserviceaccount.com", var.tuf_service_account_name, var.project_id)
+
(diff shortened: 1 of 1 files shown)


jku (approved):
Looks correct.

For completeness (assuming we don't try to find another solution) we're also going to need `roles/compute.loadBalancerAdmin` or more specifically `compute.urlMaps.invalidateCache` permission but that is not really related to the bucket so this change seems good to me.

terraform/gcp/modules/tuf/tuf.tf
```diff
@@ -119,3 +127,11 @@ resource "google_storage_bucket_iam_member" "public_tuf_preprod_member" {
...
   depends_on = [google_storage_bucket.tuf_preprod]
 }
+
+resource "google_storage_bucket_iam_member" "tuf_sa_preprod_editor" {
+  bucket = google_storage_bucket.tuf_preprod.name
+  role   = "roles/storage.objectUser"
+  member = format("serviceAccount:%s@%s.iam.gserviceaccount.com", var.tuf_service_account_name, var.project_id)
+
+  depends_on = [google_storage_bucket.tuf_preprod, google_service_account.tuf-sa]
+}
```
cpanato, line 137: nit: new line

cpanato (approved):
lgtm

haydentherapper:
> For completeness (assuming we don't try to find another solution) we're also going to need roles/compute.loadBalancerAdmin or more specifically compute.urlMaps.invalidateCache permission but that is not really related to the bucket so this change seems good to me.

Planning to grant that in the private repo since the load balancer configuration isn't in scaffolding.

Issue #948, opened 2024-01-19
tuf kms key type is incorrect


jku:
tuf/kms.tf now defines an online key for root-signing (not used yet for either prod or staging but will be). This key is EC_SIGN_P384_SHA384 which does not match the current production keytype and is in fact incompatible with cosign.

We should use EC_SIGN_P256_SHA256: this is recommended by GCP and should match the production keytype

jku:
@haydentherapper I can handle this on monday if you don't do it by then.

I'm not totally sure what the right method is though. 
* changing only the algorithm likely just changes the "algorithm for next version of this key"
* I think there's no way to manually trigger a new key version?
* Does terraform create  a completely new key if I change tuf_key_name value and the algorithm?

haydentherapper:
https://github.com/sigstore/scaffolding/pull/949 for the fix. I _think_ you can create a new key version resource to force a new version to be created in terraform.

The other option is to create a new key with a new name.

PR #949, opened 2024-01-19
Update TUF KMS key algorithm

haydentherapper:
Cosign requires ECDSA-p256 rather than p384.

Also adds protection against accidental deletion

Fixes https://github.com/sigstore/scaffolding/issues/948


haydentherapper:
Note the additional of the key version block, which will force a new key version to be created following the new template.

cpanato (approved):

bobcallaway (approved):

PR #984, opened 2024-02-16
use SA in lieu of a mapped user - argocd

bobcallaway:

### DIFF
--- a/terraform/gcp/modules/argocd/argocd.tf
+++ b/terraform/gcp/modules/argocd/argocd.tf
@@ -102,7 +102,8 @@ resource "helm_release" "argocd" {
 
   depends_on = [
     kubectl_manifest.externalsecret_argocd_ssh,
-    kubectl_manifest.externalsecret_argocd_slack
+    kubectl_manifest.externalsecret_argocd_slack,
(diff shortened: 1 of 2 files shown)


terraform/gcp/modules/argocd/argocd.tf
```diff
@@ -123,38 +124,54 @@ resource "helm_release" "argocd_apps" {
...
     name: gcp-backend
   target:
-    name: argocd-directory-api-credentials
+    name: argocd-oauth-client-credentials
   data:
-  - secretKey: googleAuth.json
+  - secretKey: client-id
     remoteRef:
-      key: "${var.gcp_secret_name_directory_api_credentials}"
+      key: "${google_secret_manager_secret.argocd-oauth-client-id.secret_id}"
```
haydentherapper, line 152: I assume this will get manually uploaded before applying terraform?

haydentherapper (approved):

terraform/gcp/modules/argocd/argocd.tf
bobcallaway: after, since terraform is creating the secret object itself.

PR #1011, opened 2024-03-06
modules/tuf: Add legacyBucketReader role for TUF SA

jku:
root-signing-staging uses "gcloud rsync" to upload files. This apparently [fails without "storage.buckets.get"](https://github.com/sigstore/root-signing-staging/issues/67)

The root cause is likely a gcloud SDK bug
(https://issuetracker.google.com/issues/323465176) but adding legacyBucketReader as a workaround seems harmless.

I'm not a terraform wizard but this "for each" mechanism seems to be used elsewhere for similar purposes.

Fixes sigstore/root-signing-staging#67

### DIFF
--- a/terraform/gcp/modules/tuf/tuf.tf
+++ b/terraform/gcp/modules/tuf/tuf.tf
@@ -75,8 +75,13 @@ resource "google_storage_bucket_iam_member" "public_tuf_member" {
 }
 
 resource "google_storage_bucket_iam_member" "tuf_sa_editor" {
+  for_each = toset([
+    "roles/storage.objectUser",
+    "roles/storage.legacyBucketReader"
+  ])
+
   bucket = google_storage_bucket.tuf.name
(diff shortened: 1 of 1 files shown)


jku:
cc @haydentherapper

terraform/gcp/modules/tuf/tuf.tf
```diff
@@ -129,8 +129,13 @@ resource "google_storage_bucket_iam_member" "public_tuf_preprod_member" {
 }
 
 resource "google_storage_bucket_iam_member" "tuf_sa_preprod_editor" {
```
jku, line 131: oops this is for preprod: this likely affects nothing in staging. I'll update this

jku: Added the same change for both preprod and prod buckets to keep things simple.

haydentherapper (approved):

PR #1062, opened 2024-04-11
Add gRPC calls for Fulcio on prober

### LINKED ISSUE #700: Add support for probing over gRPC
**Description**

We have no coverage of gRPC, and only call over HTTP. Should update https://github.com/sigstore/scaffolding/tree/main/cmd/prober

javanlacerda:
Closes #700

#### Summary

It implements a client for make requests via gRPC for prober against Fulcio. Only for reading methods were implemented.

This PR doesn't update any already existing feature, just add gRPC calls to the set of calls.


cmd/prober/prometheus.go
```diff
@@ -49,6 +49,22 @@ var (
...
+		},
+		[]string{hostLabel, methodLabel},
+	)
+
+	grpcMethodLatenciesHistogram = prometheus.NewHistogramVec(prometheus.HistogramOpts{
+		Name:    "api_endpoint_latency_histogram",
+		Help:    "API endpoint latency distribution across Rekor and Fulcio (milliseconds)",
+		Buckets: []float64{0.0, 200.0, 400.0, 600.0, 800.0, 1000.0},
+	},
+		[]string{hostLabel, methodLabel})
```
javanlacerda, line 66: Any other metrics I could add here?

cmd/prober/prober.go
```diff
@@ -201,6 +221,15 @@ func runProbers(ctx context.Context, freq int, runOnce bool) {
 				Logger.Errorf("error running request %s: %v", r.Endpoint, err)
 			}
 		}
+
+		// Performing requests against Fulcio gRPC API
+		for _, method := range fulciopb.CA_ServiceDesc.Methods {
```
haydentherapper, line 226: Rather than iterate over all possible methods, can we just pick one representative endpoint? `GetTrustBundle`?

cmd/prober/prober.go
```diff
@@ -185,7 +194,18 @@ func main() {
 	Logger.Fatal(http.ListenAndServe(addr, nil))
 }
 
-func runProbers(ctx context.Context, freq int, runOnce bool) {
+func NewFulcioClient() (fulciopb.CAClient, error) {
```
haydentherapper, line 197: nit, I would mention "GRPC" in the name since this is only creating a gRPC client. Also you can make this method lowercase so it's not exported.

cmd/prober/prober.go
```diff
@@ -170,8 +176,11 @@ func main() {
 	verificationCounter.With(prometheus.Labels{verifiedLabel: "false"}).Add(0)
 	verificationCounter.With(prometheus.Labels{verifiedLabel: "true"}).Add(0)
 
-	go runProbers(ctx, frequency, oneTime)
-
+	if fulcioClient, err := NewFulcioClient(); err != nil {
+		Logger.Errorf("error creating fulcio grpc client %v", err)
```
haydentherapper, line 180: Should this be `Fatalf` since if the client isn't properly initialized, it will fail again later?

cmd/prober/prometheus.go
haydentherapper: Can we reuse the existing metrics? The method name should be enough to distinguish, or you could also prepend `grpc://` to the host.

cmd/prober/prober.go
javanlacerda: Sure!! I tried to replicate the behavior for http requests, making requests for all available endpoints.

cmd/prober/prober.go
javanlacerda: It will not fail again. I agree that should be a Fatalf instead Error

cmd/prober/prober.go
```diff
@@ -258,6 +285,18 @@ func observeRequest(host string, r ReadProberCheck) error {
...
 
+func observeGrcpGetTrustBundleRequest(ctx context.Context, fulcioGrpcClient fulciopb.CAClient) error {
+	s := time.Now()
+
+	_, err := fulcioGrpcClient.GetTrustBundle(ctx, &fulciopb.GetTrustBundleRequest{})
+	latency := time.Since(s).Milliseconds()
+	if err != nil {
+		return err
+	}
+	exportGrpcDataToPrometheus(0, "grpc://"+fulcioGrpcURL, "GetTrustBundle", "GET", latency)
```
haydentherapper, line 296: gRPC is different than HTTP in that response status codes are a part of `error` when it returns a 4xx/5xx error. Can we log the code using something like https://github.com/sigstore/fulcio/blob/main/pkg/server/grpc_server_test.go#L126-L128 instead of `0`?

You'll need to also not return early when there is an error.

haydentherapper (approved):
Thanks!

PR #1068, opened 2024-04-18
buiding a new body for conflict reponses

### LINKED ISSUE #900: Prober retry logic can trigger Rekor failure
**Description**

We use a retryable HTTP client to make requests to Rekor (https://github.com/sigstore/scaffolding/blob/main/cmd/prober/write.go#L155). There are certain cases where a retry will always fail - For example, if the request times out, but Rekor actually succeeded in creating the entry, as the retried request will fail with an "Entry Already Exists" error.

If this error is returned, then we need to create a new blob to sign.

javanlacerda:
Closes #900

#### Summary

It adds retries updating the body for Rekor writing requests when the response has 409 code. It means that the body has conflict, then we should create a new/different body for the request.

### DIFF
--- a/cmd/prober/write.go
+++ b/cmd/prober/write.go
@@ -139,21 +139,21 @@ func fulcioWriteEndpoint(ctx context.Context, priv *ecdsa.PrivateKey) (*x509.Cer
 	return cert[0], nil
 }
 
-func makeRekorRequest(cert *x509.Certificate, priv *ecdsa.PrivateKey, hostPath string) (*http.Response, time.Time, error) {
+func makeRekorRequest(cert *x509.Certificate, priv *ecdsa.PrivateKey, hostPath string) (*http.Response, int64, error) {
 	body, err := rekorEntryRequest(cert, priv)
-	t := time.Now()
 	if err != nil {
-		return nil, t, fmt.Errorf("rekor entry: %w", err)
+		return nil, -1, fmt.Errorf("rekor entry: %w", err)
 	}
 	req, err := retryablehttp.NewRequest(http.MethodPost, hostPath, bytes.NewBuffer(body))
 	if err != nil {
-		return nil, t, fmt.Errorf("new request: %w", err)
+		return nil, -1, fmt.Errorf("new request: %w", err)
 	}
 	setHeaders(req, "")
 
-	t = time.Now()
+	t := time.Now()
 	resp, err := retryableClient.Do(req)
-	return resp, t, err
+	latency := time.Since(t).Milliseconds()
+	return resp, latency, err
 }
 
 // rekorWriteEndpoint tests the write endpoint for rekor, which is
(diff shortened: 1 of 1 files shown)


cmd/prober/write.go
```diff
@@ -151,19 +169,12 @@ func rekorWriteEndpoint(ctx context.Context, cert *x509.Certificate, priv *ecdsa
 		verificationCounter.With(prometheus.Labels{verifiedLabel: verified}).Inc()
 	}()
 
-	body, err := rekorEntryRequest(cert, priv)
-	if err != nil {
-		return fmt.Errorf("rekor entry: %w", err)
+	resp, t, err := makeRekorRequest(cert, priv, hostPath)
+	// A new body should be created when it is conflicted
+	for resp.StatusCode == http.StatusConflict {
```
javanlacerda, line 174: Does it make sense it be in a `while`? Or should I set a number of attempts ?

cmd/prober/write.go
```diff
@@ -151,24 +168,23 @@ func rekorWriteEndpoint(ctx context.Context, cert *x509.Certificate, priv *ecdsa
...
+		if resp.StatusCode != http.StatusConflict {
+			break
+		}
+		resp, t, err = makeRekorRequest(cert, priv, hostPath)
+		if err != nil {
+			return fmt.Errorf("error adding entry: %w", err)
+		}
+		defer resp.Body.Close()
+	}
+	latency := time.Since(t).Milliseconds()
```
haydentherapper, line 187: Can we record latency and response data for each request to Rekor? This would measure the total time for all requests.

cmd/prober/write.go
```diff
@@ -151,24 +168,23 @@ func rekorWriteEndpoint(ctx context.Context, cert *x509.Certificate, priv *ecdsa
...
-	t := time.Now()
-	resp, err := retryableClient.Do(req)
-	latency := time.Since(t).Milliseconds()
+	resp, t, err := makeRekorRequest(cert, priv, hostPath)
 	if err != nil {
 		return fmt.Errorf("error adding entry: %w", err)
 	}
 	defer resp.Body.Close()
+	// A new body should be created when it is conflicted
+	for {
```
haydentherapper, line 177: Can we put a reasonable limit, let's say 10, on this loop?

cmd/prober/write.go
```diff
@@ -151,24 +168,23 @@ func rekorWriteEndpoint(ctx context.Context, cert *x509.Certificate, priv *ecdsa
...
-	if err != nil {
-		return fmt.Errorf("new request: %w", err)
-	}
-
-	setHeaders(req, "")
-
-	t := time.Now()
-	resp, err := retryableClient.Do(req)
-	latency := time.Since(t).Milliseconds()
+	resp, t, err := makeRekorRequest(cert, priv, hostPath)
```
haydentherapper, line 171: Can we move this into the for loop?

cmd/prober/write.go
```diff
@@ -139,6 +139,23 @@ func fulcioWriteEndpoint(ctx context.Context, priv *ecdsa.PrivateKey) (*x509.Cer
...
+	}
+	req, err := retryablehttp.NewRequest(http.MethodPost, hostPath, bytes.NewBuffer(body))
+	if err != nil {
+		return nil, t, fmt.Errorf("new request: %w", err)
+	}
+	setHeaders(req, "")
+
+	t = time.Now()
+	resp, err := retryableClient.Do(req)
+	return resp, t, err
```
haydentherapper, line 156: I'd suggest returning the latency calculation `latency := time.Since(t).Milliseconds()` rather than the time object.

haydentherapper (changes requested):

haydentherapper (approved):
Thanks!

PR #1085, opened 2024-05-06
Add database for rekor search indexes

cmurphy:
Update the mysql and rekor modules to instantiate a new database in the primary SQL instance for search index storage.

The rekor IAM service accounts are bound to their GKE equivalents and given permission to access the Cloud SQL instance, which makes the cloud-sql-proxy sidecar in the Rekor deployment work.

The "trillian" database instance resource is renamed to "sigstore" since the instance now encompasses two databases, one of which is not for trillian.

The mysql module creates a trillian mysql user, which is not an IAM user. This user already has effectively admin grants on the SQL instance, so it is capable of connecting to the new instance and creating a new user named for the new database would not reduce the overall privileges, so we reuse the trillian mysql user for the new database.

### DIFF
--- a/terraform/gcp/modules/mysql/mysql.tf
+++ b/terraform/gcp/modules/mysql/mysql.tf
@@ -102,7 +102,7 @@ resource "random_id" "db_name_suffix" {
   byte_length = 4
 }
 
-resource "google_sql_database_instance" "trillian" {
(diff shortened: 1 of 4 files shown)


haydentherapper (approved):
Looks great!

PR #1087, opened 2024-05-10
Make terraform more configurable

cmurphy:
#### Summary

Make the sigstore terraform module more versatile for different types of deployments, especially development/proof of concept/testing deployments. Makes the following changes:

##### Make TUF bucket member configurable

In some cases it may not be desireable to expose the TUF bucket to
'allUsers', and some GCP organizations may restrict it through the
`constraints/storage.publicAccessPrevention` organization policy

##### Make DNS variables optional

For development purposes, a DNS domain may not be necessary, and
configuring one may be overkill for a short-lived proof of concept
deployment, since the registration expiration is not tied to the life of
the project.

##### Make OAuth scopes configurable

GKE applies the user.info access scope to standard clusters, in addition
to the cloud-platform scope[1]. Because of this, if the scaffolding
terraform is used to create a cluster, and then run again after
completion, the one access scope specified in terraform will be in
conflict with the two automatically assigned by GKE. Making this
configurable ensures that terraform can be idempotent if the user
provides the expected scopes for their cluster type.

[1] https://cloud.google.com/kubernetes-engine/docs/how-to/access-scopes

##### Make project_number optional

project_number is only used for monitoring, so if monitoring is not
enabled then it is not needed.

### DIFF
--- a/terraform/gcp/modules/ctlog/main.tf
+++ b/terraform/gcp/modules/ctlog/main.tf
@@ -29,9 +29,10 @@ resource "google_project_service" "service" {
 }
 
 resource "google_dns_record_set" "A_ctfe" {
-  name = "ctfe.${var.dns_domain_name}"
-  type = "A"
-  ttl  = 60
+  count = var.dns_domain_name == "" ? 0 : 1
+  name  = "ctfe.${var.dns_domain_name}"
+  type  = "A"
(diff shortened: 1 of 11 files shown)


haydentherapper (approved):
Thanks for cleaning this up!

PR #1124, opened 2024-06-01
Add playbook for redis memory increase alert

bobcallaway:
@malancas per your comment in on-call notes


malancas (approved):
Thanks for adding this

Issue #716, opened 2023-07-29
TUF: Allow for TUF to be built in a way that supports Docker/Podman not just Kubernetes


cooktheryan:
**Description**
Right now https://github.com/sigstore/scaffolding/blob/main/cmd/tuf/server/main.go is written to where the container can only run in Kubernetes. I think if we handle the initial steps to handle Kubernetes or some var if running in a container before the panic we could extend this image to be able to run in Podman or Docker.  I believe @sabre1041 had a hack for this at some point

Issue #1001, opened 2024-02-28
Support trusted root TUF target


haydentherapper:
**Description**

_Copied from https://sigstore.slack.com/archives/C049ALX6Q83/p1709072587850229_

tl;dr - Sigstore TUF metadata has evolved, but Cosign and Scaffolding are lagging behind. We need to start ingesting the new format so we can fully deprecate the old format.

For some context, a history of Sigstore TUF targets:
* Originally, TUF targets had to match expected file names, like "fulcio.crt.pem" or "ctfe.pub". You can see examples of this for [Cosign](https://github.com/sigstore/sigstore/blob/0bfcc5be818f0317a4beda7092673b4168e45da8/pkg/fulcioroots/fulcioroots.go#L38-L45). This was a problem for a lot of reasons: What happens when we need to rotate the key or cert? What happens if I want to support both a public and private deployment?
* To deal with that, we added "custom" metadata on each TUF target metadata, like https://github.com/sigstore/root-signing/blob/9f2561e6d5a8035ed2e742d3fc47907d5164baf2/repository/repository/targets.json#L41-L47. This conveyed a) what service it should be used with (Fulcio, Rekor or the CT log), b) whether it was the latest Active key or Expired (but still trusted, if untrusted it would be removed from the TUF targets), and c) the URI of the service (to support multiple instances of a log or CA in one trust root)
    * We do handle that in [Cosign](https://github.com/sigstore/sigstore/blob/0bfcc5be818f0317a4beda7092673b4168e45da8/pkg/tuf/client.go#L420-L432).
    * But this was not standardized across clients. No other Sigstore client supports custom metadata. And even within the Go clients, it's inconsistent. In [Cosign](https://github.com/sigstore/sigstore/blob/0bfcc5be818f0317a4beda7092673b4168e45da8/pkg/tuf/client.go#L116-L119), we ignore the URI, whereas in [Scaffolding](https://github.com/sigstore/scaffolding/blob/b0d09de38f7ea4ee5939a52cebadbc7127d4e127/pkg/repo/repo.go#L44-L48), we generate metadata with the URI
[...]

PR #1159, opened 2024-07-02
Enable running the TUF server outside of k8s

bkabrda:
#### Summary

This PR makes it possible to run the TUF server outside of k8s environment. Fixes https://github.com/sigstore/scaffolding/issues/716

Context: I'm a member of Red Hat's Trusted Artifact Signer product team and we're trying to make Sigstore work outside of k8s environment, in a podman-based Ansible deployment. This will allow us to start the simple TUF server in a podman pod.

#### Release Note

Made it possible to start the TUF server in a non-k8s environment.

#### Documentation

I don't believe this needs documentation, as the newly added flag is self-explanatory - but do let me know if you think this should be documented somewhere.

### DIFF
--- a/cmd/tuf/server/main.go
+++ b/cmd/tuf/server/main.go
@@ -39,27 +39,42 @@ var (
 	// root = Which holds 1.root.json
 	// repository - Compressed repo, which has been tar/gzipped.
 	secretName = flag.String("rootsecret", "tuf-root", "Name of the secret to create for the initial root file")
+	noK8s = flag.Bool("no-k8s", false, "Run in a non-k8s environment")
 )
 
-func main() {
-	flag.Parse()
+func getNamespaceAndClientset(noK8s bool) (string, *kubernetes.Clientset, error) {
+	if noK8s {
+		return "", nil, nil
+	}
 
 	ns := os.Getenv("NAMESPACE")
 	if ns == "" {
 		panic("env variable NAMESPACE must be set")
 	}
+
+	config, err := rest.InClusterConfig()
+	if err != nil {
+		return "", nil, fmt.Errorf("Failed to get InClusterConfig: %v", err)
+	}
+	clientset, err := kubernetes.NewForConfig(config)
+	if err != nil {
+		return "", nil, fmt.Errorf("Failed to get clientset: %v", err)
+	}
+
+	return ns, clientset, nil
+}
+
+func main() {
+	flag.Parse()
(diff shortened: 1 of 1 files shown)


haydentherapper:
Generally I’m ok with this, but the TUF metadata generated from scaffolding is out of date with other modern Sigstore clients. We need to complete https://github.com/sigstore/scaffolding/issues/1001. 
Id like to see that issue completed first rather than proliferate the use of this TUF repo implementation, but I also recognize there’s blocking work to get this done.

bkabrda:
Thanks for sharing the link to the trusted root TUF target issue, I wasn't aware of that and I'll definitely go through it.

My aim is to mostly use this TUF server as a "quick preview" service that a user could stand up quickly, hence I didn't dive into any other issues. I think my PR doesn't make the current situation worse, it only allows running the same code outside of k8s. Would that make it good enough to be accepted right now?

jku:
> aim is to mostly use this TUF server as a "quick preview" service that a user could stand up quickly

My only worry is that some folks will think this is a reasonable way to setup a real TUF repository. That said I don't think this patch makes things worse

bkabrda:
> My only worry is that some folks will think this is a reasonable way to setup a real TUF repository.

I totally see what you mean. I think that could perhaps be fixed by explicitly stating in the README that this is not a production-grade service and maybe also emitting a warning logline saying this when starting the TUF server?

bkabrda:
@haydentherapper hi, is there anything specific I can do to help get this PR reviewed? Or should I consider it blocked on the issue you linked?

haydentherapper:
@bkabrda I'm OK to merge this, I don't want to block y'all, but if you are interested, we'd love some help pushing forward updates in scaffolding around generating modern TUF repos.

haydentherapper (approved):

bkabrda:
@haydentherapper thanks! I was just thinking about the improvements and I think I have a reasonable proposal. I'll open a separate issue to discuss it and CC you on it. I'm happy to work on the implementation of the proposal assuming it will look ok to you.

bkabrda:
I opened the proposal for improvements [here](https://github.com/sigstore/scaffolding/issues/1182). It took a little longer than I promised as I was caught up in other things... Anyway, I'd love to hear thoughts on the proposal from both of you - and if it looks good, I would really appreciate if we could get this PR approved and merged. Thanks!

PR #1160, opened 2024-07-02
Add v1 Fulcio endpoint to prober

cmurphy:
There is an SLO set up for the /api/v1/signingCert Fulcio endpoint[1], but it is currently reporting "No SLO status data" because the prober was never testing that endpoint. This lead to an outage that went undetected by the monitoring system.

Cosign uses the legacy certificate request endpoint in its Fulcio client[2][3]. This means that the v1 endpoint is likely the most used and therefore an important health indicator. This change adds the v1 endpoint to the prober test, which should populate Prometheus with data which should activate the SLO.

[1] https://github.com/sigstore/scaffolding/blob/8f7aa097e54eabcecbc671818f9eb5f0e723e54b/terraform/gcp/modules/monitoring/fulcio/slo.tf#L79-L83
[2] https://github.com/sigstore/cosign/blob/79db196e2d97e7dfc4d8201ef829d4ce906605a7/cmd/cosign/cli/fulcio/fulcio.go#L32
[3] https://github.com/sigstore/fulcio/blob/07b19da442b418ebcf072ac65a7abb25f0e3d5c8/pkg/api/client.go#L60

### DIFF
--- a/cmd/prober/write.go
+++ b/cmd/prober/write.go
@@ -24,6 +24,7 @@ import (
 	"crypto/x509"
 	"encoding/hex"
 	"encoding/json"
+	"encoding/pem"
 	"fmt"
 	"io"
 	"net/http"
@@ -52,8 +53,9 @@ const (
 	defaultOIDCIssuer   = "https://oauth2.sigstore.dev/auth"
 	defaultOIDCClientID = "sigstore"
 
-	fulcioEndpoint = "/api/v2/signingCert"
-	rekorEndpoint  = "/api/v1/log/entries"
+	fulcioEndpoint       = "/api/v2/signingCert"
+	fulcioLegacyEndpoint = "/api/v1/signingCert"
+	rekorEndpoint        = "/api/v1/log/entries"
 )
 
 func setHeaders(req *retryablehttp.Request, token string) {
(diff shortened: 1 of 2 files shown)


cmd/prober/write.go
```diff
@@ -68,8 +70,75 @@ func setHeaders(req *retryablehttp.Request, token string) {
...
+	responseBody, err := io.ReadAll(resp.Body)
+	if err != nil {
+		Logger.Errorf("error reading response from Fulcio: %v", err)
+		return nil, err
+	}
+	certBlock, chainPEM := pem.Decode(responseBody)
+	if certBlock == nil || chainPEM == nil {
+		Logger.Errorf("did not find expected certificates")
+	}
+	intermediateBlock, rootPEM := pem.Decode(chainPEM)
```
haydentherapper, line 121: Just so this can support probing custom instances with any number of intermediates, we might want to continually decode until `pem.Decode()` parses the root and `rest` is empty, or use `UnmarshalCertificatesFromPEM(chainPEM)`. We could add a configuration flag for how many intermediates are expected?

haydentherapper (commented):
I wonder if this was broken since https://github.com/sigstore/scaffolding/pull/540...

cmd/prober/write.go
cmurphy: I wrote it this way to be consistent with the v2 test which expects exactly 3: https://github.com/sigstore/scaffolding/blob/f775933430a4f3b3946595f0e7dccef75e16309b/cmd/prober/write.go#L121-L125

I think making it configurable would be a separate scope.

haydentherapper: Ah fair point, then this LGTM!

haydentherapper (approved):

cmurphy:
Accidentally pushed a commit to the wrong branch

haydentherapper (approved):

PR #1174, opened 2024-07-12
Add documentation on generating trusted_root.json

cmurphy:
The TUF mirror in this scaffolding stack does not serve a trusted_root.json, which is needed by sigstore-go to verify artifacts signed using a private sigstore deployment. Add documentation so that users can create this themselves using available community tooling. This may eventually be turned into part of the GitHub action so that clients can use the action for integration testing.

I couldn't find an official way of generating this file that felt safe to include as part of the [createsecret](https://github.com/sigstore/scaffolding/blob/main/cmd/tuf/createsecret/main.go) service, but I still wanted to write this down so others don't fall down the same rabbit hole wondering why a client needs this apparently missing file. I used this to generate the trusted material for [this PR](https://github.com/sigstore/sigstore-go/pull/229) because I needed to test the changes with a custom OIDC provider.

### DIFF
--- a/getting-started.md
+++ b/getting-started.md
@@ -384,3 +384,33 @@ vaikas@villes-mbp scaffolding % echo 'eyJfdHlwZSI6Imh0dHBzOi8vaW4tdG90by5pby9TdG
 ```
 
 Notice our predicate is `foobar test attestation` as was in our predicate file.
+
+## Generating trusted_root.json
+
+The TUF mirror in this stack does not serve a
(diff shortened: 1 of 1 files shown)


kommendorkapten (approved):
Nice!

jku (approved):
I've never modified trusted_root.json myself but this seems reasonable

haydentherapper (approved):
Thanks!

Issue #1182, opened 2024-07-16
Proposal: Make the TUF server more useful for production deployment


bkabrda:
**Description**

Hi folks, as a followup to the discussion that happened in https://github.com/sigstore/scaffolding/pull/1159, I wanted to make the following proposal to improve the TUF server. I would be curious what are your thoughts; I'd definitely love to work on this myself to push things forward.

**Desired state**

I would like to make the TUF server usable in production environments. My aim is to support both k8s and non-k8s environments (as noted in https://github.com/sigstore/scaffolding/pull/1159). To achieve that, there are several things that I think should happen when the TUF server starts:
* Check if there is a complete TUF repository to serve. If so, it would just serve that.
* If there is no TUF repository to serve, it would do what it does now - scaffold the TUF repository and start serving it.
  * But it would also store all the generated private keys, so that the user can retrieve them and use them to continue operating the TUF repository outside of the TUF server.

This would mean we:
* Have a TUF server that has a quick no-touch start (as it does right now).
* Don't reset the TUF repository on a pod relocation.
* Have a way for the user to grab the generated TUF repository and continue operating it (add more keys, etc).
* Have a way for the user to pre-create the TUF repository and the TUF server would just serve it without creating anything itself.

**Implementation details**

The implementation would be very similar for both k8s and non-k8s environments. Roughly speaking:
* Currently, the TUF server generates a TUF repository and stores it in a secret. We could have that secret mounted inside the TUF pod (if it exists) and serve the TUF root from there.
  * Non-k8s case: assume the repository is in a directory on the filesystem and serve from there (again, if it exists).
* If there is no TUF repository to serve:
  * Generate a new one and store it in the secret just like we do now.
    * Non-k8s case: store it in a directory on the filesystem.
  * Also store the private keys in a different secret for the user to grab them.
    * Non-k8s case: store them in a different directory on the filesystem.
* Additionally, we could have a switch in the server executable which would specify whether the server should generate only the "new-style" trust root (trusted_root.json) or also include the current "old-style" TUF targets.

CC @haydentherapper @jku

bkabrda:
Thanks a lot for providing the context!

[...]

PR #1191, opened 2024-07-24
Add functionality to generate trusted_root.json by the TUF server

bkabrda:
#### Summary

This PR is the first of a series to address https://github.com/sigstore/scaffolding/issues/1182. It adds functionality to the TUF server to generate a `trusted_root.json` as a target.

There is some overlap with the [trtool](https://github.com/kommendorkapten/trtool) project and in the future, ideally we would make a library to pull out the common functionality and share it.

#### Release Note

The TUF server now generates a `trusted_root.json` target from the supplied files.

#### Documentation

I don't believe that the TUF server is even documented anywhere, so I don't think documentation change is required, but please let me know if I'm wrong about that.

### DIFF
--- a/pkg/repo/repo.go
+++ b/pkg/repo/repo.go
@@ -18,7 +18,14 @@ import (
 	"archive/tar"
 	"compress/gzip"
 	"context"
+	"crypto/ecdsa"
+	"crypto/ed25519"
+	"crypto/elliptic"
+	"crypto/rsa"
+	"crypto/sha256"
+	"crypto/x509"
 	"encoding/json"
+	"encoding/pem"
 	"errors"
 	"fmt"
 	"io"
@@ -26,12 +33,30 @@ import (
 	"os"
 	"path"
 	"path/filepath"
+	"sort"
 	"strings"
 	"time"
 
+	v1Common "github.com/sigstore/protobuf-specs/gen/pb-go/common/v1"
+	v1 "github.com/sigstore/protobuf-specs/gen/pb-go/trustroot/v1"
 	"github.com/theupdateframework/go-tuf"
+	"google.golang.org/protobuf/encoding/protojson"
+	"google.golang.org/protobuf/types/known/timestamppb"
 	"knative.dev/pkg/logging"
 )
+
+const (
+	FulcioTarget  = "Fulcio"
+	RekorTarget   = "Rekor"
+	CTFETarget    = "CTFE"
+	TSATarget     = "TSA"
+	UnknownTarget = "Unknown"
+)
+
+type CreateRepoOptions struct {
+	AddMetadataTargets bool
+	AddTrustedRoot     bool
+}
 
 // TargetWithMetadata describes a TUF target with the given Name, Bytes, and
 // CustomMetadata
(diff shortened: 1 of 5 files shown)


pkg/repo/repo.go
```diff
@@ -127,36 +159,270 @@ func CreateRepoWithMetadata(ctx context.Context, targets []TargetWithMetadata) (
...
+		if err != nil {
+			return nil, "", fmt.Errorf("failed to construct trust root: %w", err)
+		}
+		targets = append(targets, *trustedRootTarget)
+	}
+
 	return CreateRepoWithMetadata(ctx, targets)
 }
 
+func constructTrustedRoot(targets []TargetWithMetadata) (*TargetWithMetadata, error) {
```
haydentherapper, line 199: Can we move this to sigstore-go? There is some overlap with https://github.com/sigstore/cosign/issues/3794, and I'd prefer to have this centrally located right off the bat.

@cmurphy FYI, since you were asking about utilities for trust root generation as well.

pkg/repo/repo.go
```diff
@@ -127,36 +159,270 @@ func CreateRepoWithMetadata(ctx context.Context, targets []TargetWithMetadata) (
...
+		if v.Curve == elliptic.P384() {
+			return v1Common.PublicKeyDetails_PKIX_ECDSA_P384_SHA_384, nil
+		}
+		if v.Curve == elliptic.P521() {
+			return v1Common.PublicKeyDetails_PKIX_ECDSA_P521_SHA_512, nil
+		}
+		return 0, errors.New("unsupported elliptic curve")
+	case *rsa.PublicKey:
+		/*
+			NOTE: It is not possible to recognize padding from just the public key alone;
```
haydentherapper, line 343: I'd just omit this commented out block, almost all Sigstore clients can't handle RSA-PSS keys.

pkg/repo/repo.go
```diff
@@ -127,36 +159,270 @@ func CreateRepoWithMetadata(ctx context.Context, targets []TargetWithMetadata) (
...
+		return targets[i].Name < targets[j].Name
+	})
+
+	for _, target := range targets {
+		// NOTE: in the below switch, we are able to process whole certificate chains, but we also support
+		// if they're passed in as individual certificates, already split in individual targets
+		switch getTargetUsage(target.Name) {
+		case FulcioTarget:
+			switch {
+			case strings.Contains(target.Name, "leaf"):
```
haydentherapper, line 219: There should be no leaves in the trust root for Fulcio, it should only be intermediates and roots. The leaf would be the code signing certificate.

pkg/repo/repo.go
```diff
@@ -127,36 +159,270 @@ func CreateRepoWithMetadata(ctx context.Context, targets []TargetWithMetadata) (
...
+	}
+}
+
+func certChainToAuthority(certChainPem []byte) (*v1.CertificateAuthority, error) {
+	var cert *x509.Certificate
+	var err error
+	rest := certChainPem
+	certChain := v1Common.X509CertificateChain{Certificates: []*v1Common.X509Certificate{}}
+
+	// skip potential whitespace at end of file (8 is kinda random, but seems to work fine)
```
haydentherapper, line 381: I'm not sure this is necessary, I believe whitespace will be handled by `pem.Decode`. Would be good to test for that.

haydentherapper (changes requested):
This is great, thank you!

Just left a few comments. One other one is that I think this is useful as a library, so can we add it straight to sigstore-go?

bkabrda:
Hi, thanks for the feedback! I can certainly try to add this to sigstore-go. I'll leave this PR open and when the sigstore-go work is merged, I'll update it to use that as a library from here.

bkabrda:
Oh, I didn't realize that sigstore-go already has so many related code to this in https://github.com/sigstore/sigstore-go/tree/main/pkg/root, this actually makes so much sense... I don't know why I didn't think of looking there first. Thanks again for pointing me in the right direction!

bkabrda:
Hi :wave: I created https://github.com/sigstore/sigstore-go/pull/247. I have a local modification of this PR that uses the functionality from there and works fine. I'll push it once the sigstore-go PR is merged (ideally we would also release that if possible, not sure if/how I could help with that).

bkabrda:
This PR is conflicting on multiple files and will need a significant rewrite to utilize https://github.com/sigstore/sigstore-go/pull/247 anyway. I think it's going to be easier if I close it and open a new one instead.

PR #1214, opened 2024-08-08
Add different run modes for the TUF server, allow saving TUF keys as a secret

bkabrda:
#### Summary

Addresses part of https://github.com/sigstore/scaffolding/issues/1182.

This commit implements features necessary to run/operate the TUF server in production much better:

* The TUF server can now be run in 4 different modes:
  * `init` - only init the TUF repository and exit
  * `init-no-overwrite` - same as `init`, but won't overwrite the TUF repository if it already exists
  * `serve` - only serve an existing TUF repository (no `init`)
  * `init-and-serve` - `init` and then `serve` (this is default to preserve backwards compatibility)
* The TUF repository can now be initialized/served from a given path, as opposed to always living in a `/tmp` directory.
* The TUF keys can optionally be exported as an individual k8s secret.

Some notes:
* The idea behind this PR is that:
  * There is a k8s Job/initContainer that runs the server with either `init` or `init-no-overwrite` to create the TUF repository at a mounted volume.
  * TUF server Deployment starts using only `serve` to serve the TUF repository from the mounted volume.
  * TUF keys are saved in the secret, so the admin of the system can download them and keep operating the TUF repository - they do so by updating TUF files localy and placing them in the volume.
* The PR could optionally be split into multiple PRs. I think this functionality is interconnected and hence makes sense to submit as a single PR, but let me know if you want me to split it into multiple ones.
* The choice to save the TUF keys into an individual secret (as opposed to adding them to the one that is already created) is kind of arbitrary. We could very well put everything in one secret, however I think having the option to delete the secret with keys (after it's downloaded by the admin of the system) and keeping the one with the repository is good to have.

#### Release Note

* The TUF server was modified to be run in 4 different modes:
  * `init` - only init the TUF repository and exit
  * `init-no-overwrite` - same as `init`, but won't overwrite the TUF repository if it already exists
  * `serve` - only serve an existing TUF repository (no `init`)
  * `init-and-serve` - `init` and then `serve` (this is default to preserve backwards compatibility)
* The TUF repository was made to be initialized/served from a given path, as opposed to always living in a `/tmp` directory.
* Option to export the TUF keys an individual k8s secret was added.

#### Documentation

It feels like this change would deserve some docs, but I don't know if there even are docs for the TUF server. I can certainly write something up if you can point me to the right place.


jku (approved):
I think this makes sense: it makes it clearer that this is really just file serving code with a single function to (now optionally) initialize that file server with some generated TUF content.


I don't comment on the secrets management: I don't know this technology at all

cmd/tuf/server/main.go
```diff
@@ -140,26 +149,112 @@ func main() {
...
 		nsSecret := clientset.CoreV1().Secrets(ns)
-		if err := secret.ReconcileSecret(ctx, *secretName, ns, data, nsSecret); err != nil {
-			logging.FromContext(ctx).Panicf("Failed to reconcile secret %s/%s: %v", ns, *secretName, err)
+		if err := secret.ReconcileSecret(ctx, repoSecretName, ns, data, nsSecret); err != nil {
+			return fmt.Errorf("failed to reconcile secret %s/%s: %v", ns, repoSecretName, err)
+		}
+
+		// If we should also store created keys in a secret, read all their files and save them in the secret
+		if keysSecretName != "" {
+			keyFiles, err := os.ReadDir(filepath.Join(dir, "keys"))
```
evankanderson, line 164: Didn't we pack the `keys` directory into `data["repository"]` already, along with the rest of `dir`?

cmd/tuf/server/main.go
```diff
@@ -140,26 +149,112 @@ func main() {
...
+				return fmt.Errorf("failed to reconcile keys secret %s/%s: %v", ns, keysSecretName, err)
+			}
+		}
+	}
+
+	logging.FromContext(ctx).Infof("tuf repository was created in: %s, moving to %s", dir, targetDir)
+
+	// Copy repository to the targetDir - until Go 1.23 which has os.CopyFS, we use
+	// a quick hack where we uncompress the compressed repository to the targetDir
+	repo.Uncompress(bytes.NewReader(data["repository"]), targetDir)
```
evankanderson, line 190: Question: Right now, it seems like we're creating in a temp directory and then moving to the final destination.  What about having the `repo.CreateRepo` take `targetDir` as an argument, and not needing to do this unpack?

At that point, the code could also look like:

```golang
if noK8s {
  // We're done
  return nil
}
```

I suppose one advantage of the copy approach is that you won't end up with a half-initialized directory if you crash partway through....

evankanderson (commented):
@haydentherapper asked me to look at this.  The k8s part makes sense, but I had a few questions about the general patterns since I'm not as familiar with the codebase.

Splitting `init` and `serve` makes a lot of sense to me.  I _am_ wondering about the "saving the keys" vs simply pulling the keys out of the existing saved secret.

cmd/tuf/server/main.go
```diff
@@ -33,13 +34,24 @@ import (
...
-	dir = flag.String("file-dir", "/var/run/tuf-secrets", "Directory where all the files that need to be added to TUF root live. File names are used to as targets.")
+	dir       = flag.String("file-dir", "/var/run/tuf-secrets", "Directory where all the files that need to be added to TUF root live. File names are used to as targets.")
+	targetDir = flag.String("target-dir", "", "Directory where TUF repository should be created/served from. Defaults to temporary directory.")
+	mode      = flag.String("mode", modeInitAndServe, "Run mode of the TUF server. One of: init, init-and-serve, serve, init-no-overwrite")
 	// Name of the "secret" where we create two entries, one for:
 	// root = Which holds 1.root.json
 	// repository - Compressed repo, which has been tar/gzipped.
 	secretName = flag.String("rootsecret", "tuf-root", "Name of the secret to create for the initial root file")
-	noK8s = flag.Bool("no-k8s", false, "Run in a non-k8s environment")
+	// Name of the "secret" where we create one entry "keys" which is a compressed directory of private keys generated by TUF
```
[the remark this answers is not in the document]
bkabrda: Oh, this comment is actually no longer true. It was true in the first version of my code, but now it is not tar/gzipped, but it's literally a secret with key/value pairs for the 4 individual keys. I will update the comment.

cmd/tuf/server/main.go
```diff
@@ -64,24 +76,20 @@ func getNamespaceAndClientset(noK8s bool) (string, *kubernetes.Clientset, error)
...
 	}
 
-	tufFiles, err := os.ReadDir(*dir)
+	tufFiles, err := os.ReadDir(certsDir)
 	if err != nil {
-		logging.FromContext(ctx).Fatalf("failed to read dir %s: %v", *dir, err)
+		return fmt.Errorf("failed to read dir %s: %v", certsDir, err)
 	}
-	trimDir := strings.TrimSuffix(*dir, "/")
+	trimDir := strings.TrimSuffix(certsDir, "/")
```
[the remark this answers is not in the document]
bkabrda: Sounds good, I can do that.

cmd/tuf/server/main.go
bkabrda: We didn't. If you look at how this is done:

```
if err := repo.CompressFS(os.DirFS(dir), &compressed, map[string]bool{"keys": true, "staged": true}); err != nil
```

and how the `CompressFS` function works, it will actually intentionally skip the `keys` and `staged` directories because of how it is called.

cmd/tuf/server/main.go
bkabrda: There were two reasons why I didn't add the extra argument, even though we considered it:

* Exactly as you said, I think it's good to not end up with a potentially half-initialized directory.
* Adding the new argument would change the public API of `repo.CreateRepo`. While I think that the modules from this repo might not be depended on very much, I wanted to prevent API breakage (but actually talking about this right now, I'm curious if that's ok - for future contributions).

bkabrda:
@evankanderson thanks a lot for your review! I think I addressed all the points. As noted in one of the inline comments, the currently saved secret intentionally explicitly excludes the keys directory from being saved. I'm not sure why it was coded this way, but I think it makes sense that these two are independent and the keys secret can be downloaded and deleted altogether. (I don't think this is a huge benefit, but it felt to me that it might be nice, YMMV of course).

evankanderson (approved):

PR #1278, opened 2024-09-23
update vendor

k4leung4:


haydentherapper:
Would it be possible to split this into PRs updating the vendored filed and them go 1.23, in case either causes issues?

bobcallaway (approved):

PR #1333, opened 2024-11-06
change in-cluster prober to read across all shards (programatically)

bobcallaway:
Dynamically pulls the list of shards from Rekor, chooses a random entry within each shard, and reads those entries. This provides us better operational coverage given each shard is persisted as a different tree.


cpanato (approved):

haydentherapper (commented):
Late LGTM

PR #1392, opened 2024-12-21
move SQL for rekor indices into rekor module, add cloud sql iam user

bobcallaway:

### DIFF
--- a/terraform/gcp/modules/sigstore/sigstore.tf
+++ b/terraform/gcp/modules/sigstore/sigstore.tf
@@ -200,6 +200,10 @@ module "mysql" {
   ]
 }
 
+moved {
+  from = module.mysql.google_sql_database.searchindexes
+  to   = module.rekor.google_sql_database.searchindexes
+}
 
 // Rekor
 module "rekor" {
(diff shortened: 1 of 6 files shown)


hectorj2f (approved):

terraform/gcp/modules/sigstore/sigstore.tf
```diff
@@ -231,6 +235,8 @@ module "rekor" {
 
   redis_cluster_memory_size_gb = var.redis_cluster_memory_size_gb
 
+  index_database_instance_name = module.mysql.mysql_instance
+
   depends_on = [
```
haydentherapper, line 240: Do you need to add `mysql.mysql_instance` for a dependency?

bobcallaway: Since this is an [output of the mysql module](https://github.com/sigstore/scaffolding/blob/9abab05d155fa0ae8349c234eb8a609db161cbf0/terraform/gcp/modules/mysql/outputs.tf#L23), the dependency is implicitly created.

haydentherapper (approved):

PR #1404, opened 2025-01-06
Remove Cloud SQL memory utilization > 90% alert

priyawadhwa:
We've been getting a lot of low urgency pages for this alert (~50 in the past day). As discussed in the oncall meeting, this isn't a particularly useful  metric to alert on since MySQL will always use a large amount of memory. CPU usage has been normal, and we have alerts for it, and it's probably a better metric to alert on


haydentherapper (commented):
Do you think we should keep this but bump it far higher, to 95% or 99? I'd worry having no alert might miss a memory leak.

priyawadhwa:
There's an alert for 95% right below this one in the file!

haydentherapper (approved):
NVM, thank you!

PR #1448, opened 2025-02-03
add automation to update ko.yaml for csp updates

bobcallaway:
this adds some automation to ensure `.ko.yaml` gets updated automatically when a dependabot update is proposed in `Dockerfile.deps`


cpanato (approved):
Nice

PR #1485, opened 2025-02-27
workflows: Test the generated trusted_root.json

jku:
Use cosign to verify that the trust root generated by scaffolding actually works: this currently requires specifying --trusted-root when calling cosign

**DRAFT**: I expect the new test to fail until I add the fix


jku:
oh. I may have misunderstood what that whole workflow is testing: it always installs scaffolding from main... I'm not sure what this is supposed to do on PRs?

jku:
I'm testing a simplified setup script in the workflow itself: if this does not work, I think I'll give up fixing the tests

jku:
Closing this : test-action-tuf is clearly not a good place for this test

PR #1486, opened 2025-02-27
Fix scaffolding trust root

jku:
Scaffolding TUF setup includes a generated trusted_root.json but it is never tested and manual testing implies it is broken
* Add test to verify the issue
* Fix the LogId issue in trusted_root.json


cmurphy (approved):

haydentherapper (approved):
Thanks!

PR #1517, opened 2025-03-17
feat: add portable test containers

ramonpetgrave64:
https://github.com/sigstore/rekor-tiles/issues/38

#### Summary

Adds a new script for making it easier to launch a local fulcio and rekor for testing, including within Github Actions.

## Testing

currently testing in a pending cosign PR https://github.com/sigstore/cosign/actions/runs/13926226723/job/38971336997?pr=4117

#### Documentation

Add Makefile and action for launching a local fulcio and rekor.

# Followup PRs:

- [ ] install docker to macos runner in CI test: https://github.com/GoogleContainerTools/container-structure-test/blob/56c7201716d770c0f820a9c19207ba2ea77c34f8/.github/workflows/integration-tests.yml#L24-L30

### DIFF
--- a/.github/workflows/test-portable.yaml
+++ b/.github/workflows/test-portable.yaml
@@ -13,7 +13,7 @@ jobs:
       matrix:
         platform:
           - ubuntu-latest
-          - macos-13
+          # - macos-13 # not yet working
           # - macos-latest # docker not supported yet. See https://github.com/marketplace/actions/setup-docker-on-macos#arm64-processors-m1-m2-m3-series-used-on-macos-14-images-are-unsupported.
           # - windows-latest # hold off while rekor still specifies `platform: linux`
         rekor-repo:
--- a/.github/workflows/test-portable.yaml
+++ b/.github/workflows/test-portable.yaml
@@ -8,6 +8,8 @@ jobs:
   test:
     strategy:
       fail-fast: false
+      # Jobs seem to run more slowly when run concurrently.
+      max-parallel: 1
       matrix:
         platform:
           - ubuntu-latest
--- a/.github/workflows/test-portable.yaml
+++ b/.github/workflows/test-portable.yaml
@@ -1,0 +1,31 @@
+name: Portable Test Sigstore Containers
+on:
+  workflow_dispatch:
+  push:
+  pull_request:
+
+jobs:
+  test:
+    strategy:
+      fail-fast: false
+      matrix:
+        platform:
+          - ubuntu-latest
+          - macos-13
+          # - macos-latest # docker not supported yet. See https://github.com/marketplace/actions/setup-docker-on-macos#arm64-processors-m1-m2-m3-series-used-on-macos-14-images-are-unsupported.
+          # - windows-latest # hold off while rekor still specifies `platform: linux`
+        rekor-repo:
+          - https://github.com/sigstore/rekor.git
+          - https://github.com/sigstore/rekor-tiles.git
+    runs-on: ${{ matrix.platform }}
+    steps:
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+        with:
+          persist-credentials: false
+      - uses: ./actions/portable_testing
+        with:
+          rekor-repo: ${{ matrix.rekor-repo }}
+      - name: fulcio healthcheck
+        run: curl http://localhost:5555/healthz
+      - name: rekor healthcheck
+        run: curl http://localhost:3000/
--- a/actions/portable_testing/Makefile
+++ b/actions/portable_testing/Makefile
@@ -1,0 +1,71 @@
+# Copyright 2025 The Sigstore Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# optionally set these env variables
+FULCIO_REPO ?= https://github.com/sigstore/fulcio.git
+REKOR_REPO ?= https://github.com/sigstore/rekor-tiles.git
+
+# prevent a conflict with rekor-server's metrics port.
+export FULCIO_METRICS_PORT = 2113
+
+CHECKOUT_DIR ?= ./CHECKOUT_DIR
+FULCIO_DIR ?= $(CHECKOUT_DIR)/fulcio
+REKOR_DIR ?= $(CHECKOUT_DIR)/rekor
+
+.PHONY: all
+all: checkout up
+
+.PHONY: up
+up:
+# bring up both fulcio and rekor simultaneously.
+	make up-fulcio up-rekor -j 2
+
+.PHONY: down
+down:
+# bring down both fulcio and rekor simultaneously.
+	make down-fulcio down-rekor -j 2
+
+.PHONY: up-fulcio
+up-fulcio:
+	cd $(FULCIO_DIR) && docker compose up --wait
+
+.PHONY: down-fulcio
+down-fulcio:
+	cd $(FULCIO_DIR) && docker compose down
+
+.PHONY: up-rekor
+up-rekor:
+	cd $(REKOR_DIR) && docker compose up --wait
+
+.PHONY: down-rekor
+down-rekor:
+	cd $(REKOR_DIR) && docker compose down
+
+.PHONY: checkout
+checkout:
+# checkout the repos locally.
+ifeq ($(wildcard $(FULCIO_DIR)),)
+	git clone --depth 1 $(FULCIO_REPO) $(FULCIO_DIR)
+else
+	@echo "Already cloned fulcio"
+endif
+ifeq ($(wildcard $(REKOR_DIR)),)
+	git clone --depth 1 $(REKOR_REPO) $(REKOR_DIR)
+else
+	@echo "Already cloned rekor"
+endif
+
+.PHONY: clean
+clean:
+	rm -rf $(FULCIO_DIR) $(REKOR_DIR)
--- a/actions/portable_testing/Makefile
+++ b/actions/portable_testing/Makefile
@@ -23,34 +23,10 @@ CHECKOUT_DIR ?= ./CHECKOUT_DIR
 FULCIO_DIR ?= $(CHECKOUT_DIR)/fulcio
 REKOR_DIR ?= $(CHECKOUT_DIR)/rekor
 
+# all
+
 .PHONY: all
-all: checkout up
-
-.PHONY: up
-up:
-# bring up both fulcio and rekor simultaneously.
-	make up-fulcio up-rekor -j 2
-
-.PHONY: down
-down:
-# bring down both fulcio and rekor simultaneously.
-	make down-fulcio down-rekor -j 2
-
-.PHONY: up-fulcio
-up-fulcio:
-	cd $(FULCIO_DIR) && docker compose up --wait
-
-.PHONY: down-fulcio
-down-fulcio:
-	cd $(FULCIO_DIR) && docker compose down
-
-.PHONY: up-rekor
-up-rekor:
-	cd $(REKOR_DIR) && docker compose up --wait
-
-.PHONY: down-rekor
-down-rekor:
-	cd $(REKOR_DIR) && docker compose down
+all: checkout build up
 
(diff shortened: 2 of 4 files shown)


ramonpetgrave64:
@jku @cmurphy @loosebazooka

.github/workflows/test-portable.yaml
```diff
@@ -0,0 +1,24 @@
...
+    runs-on: ${{ matrix.platform }}
+    steps:
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+        with:
+          persist-credentials: false
+      - uses: ./actions/portable_testing
+      - name: healthcheck
+        run: |
+          curl http://localhost:5555/healthz
+          curl http://localhost:3000/
```
jku, line 24: this could really use a comment (or maybe just a more descriptive name)

jku (commented):
Can you add a PR description or maybe a note in the README that explains why this exists: when is it better than `actions/setup` + `hack/setup-scaffolding*` (and vice versa)?

Is there a reason that the cosign PR does not use the action?

actions/portable_testing/action.yml
```diff
@@ -0,0 +1,43 @@
...
+        path: |
+          ~/.cache/go-build
+          ~/go/pkg/mod
+
+          ~/Library/Caches/go-build
+          ~/go/pkg/mod
+
+          ~\AppData\Local\go-build
+          ~\go\pkg\mod
+        key: ${{ runner.os }}-go-${{ hashFiles('**/go.sum') }}
```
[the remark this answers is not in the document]
haydentherapper: Given this is building with docker, why is this cache needed?

haydentherapper (commented):
This should all services needed to test a Sigstore client, including the timestamp authority, an OIDC identity provider, and a trusted root file to be used by the client.

actions/portable_testing/Makefile
```diff
@@ -0,0 +1,71 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# optionally set these env variables
+FULCIO_REPO ?= https://github.com/sigstore/fulcio.git
+REKOR_REPO ?= https://github.com/sigstore/rekor.git
```
cmurphy, line 17: The main reason we want this action is to test rekor-tiles, can we go ahead and use that here? Or make it configurable?

actions/portable_testing/action.yml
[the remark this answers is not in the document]
ramonpetgrave64: Both the cloned docker-compose.yml files will do a `go build` of their services, so this caching saves time when the `go build` starts downloading the go module dependencies.

We are using docker to build images locally within the Actions run, but since I haven't yet figured out how to cache image layers, the cache for the go files still offers a good speedup.

actions/portable_testing/Makefile
ramonpetgrave64: Eventually yes, but iiuc rekor-tiles has a slightly different APIs that clients like cosign are not yet ready to use.

cmurphy: But they don't have to use this, they can use what they already have. They need this in order to update.

ramonpetgrave64: Ok I see what you mean: Have this Makefile is rekor-tiles as default, and clients may optionally use rekor-v1 by setting the REKOR_REPO env variable. Right?

cmurphy: Yes. But I think many clients may not bother to use this for rekor v1 because they already have their own ways of testing against rekor v1.

actions/portable_testing/action.yml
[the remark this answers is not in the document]
jku: right but when e.g. test-portable.yaml runs this action, actions/cache does `hashFiles('**/go.sum')` before anything has been checked out.

.github/workflows/test-portable.yaml
ramonpetgrave64: It seemed like it was faster, but I can see now that the caching might not actually be happening. I'm trying fix this.

- https://github.com/sigstore/scaffolding/actions/runs/13926652632/job/38972863790#step:7:3

```
Warning: Path Validation Error: Path(s) specified in the action for caching do(es) not exist, hence no cache is being saved.
```

actions/portable_testing/action.yml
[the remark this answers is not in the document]
haydentherapper: I might be misunderstanding the interaction between Docker and the cache, but the Docker-managed build happens in a container that does not have access to the GitHub Action cache.

https://docs.docker.com/build/ci/github-actions/cache/#cache-mounts mentions a solution, which would make the cache directory accessible to the Docker container.

https://docs.docker.com/build/cache/backends/gha/#using-dockerbuild-push-action describes how to cache layers. Either of these approaches seem usable.

I'm also fine with omitting caching for now, it adds complexity and might not be that significant of a time save.

jku:
> > Can you add a PR description or maybe a note in the README that explains why this exists: when is it better than `actions/setup` + `hack/setup-scaffolding*` (and vice versa)?
> > Is there a reason that the cosign PR does not use the action?
> 
> Added to hte cosing PR's description "we could use the Action, but we still want to be able to do all this locally, so we use the Makefile directly."

So why build the action at all then -- who is it for?

haydentherapper:
@jku The purpose of https://github.com/sigstore/rekor-tiles/issues/38 is to create a single script for testing clients against real Sigstore services. The problem now is that we have three divergent solutions - the Cosign script that spins up local services with Docker along with some services in the Go e2e test (TUF and the TSA), Scaffolding which spins up a Kubernetes cluster, and the sigstore-js service mocks (which is used by sigstore-ruby too).

The problems now:

* For Scaffolding, Kubernetes is overkill for both CI and local development. Also, the Kubernetes configs (not maintained) have very much diverged from the Helm charts which are actively maintained (A side effect of this work will be to propose deprecating the Kubernetes configs, as we still see blog posts recommending deploying local Sigstore from Scaffolding, but that's not directly coupled to Rekor v2)
* Also for Scaffolding, we have a mix of useful binaries, like the tooling to create a local TUF repo and trust bundle, but these aren't used by CI without also using the Kubernetes stack.
* The sigstore-js mocks are mocks and not real services.
* For Cosign, some infrastructure is spun up via Docker, some within the Go test runner. This isn't a significant issue, just makes it not portable for CI testing in non-Go clients.

Cosign is the most up to date testing solution and what I had recommended we build on. For the script, it should:

1. Generate key material
2. `docker compose up` for Rekor, Fulcio, and the TSA
3. Spin up the fake OIDC server that we have in Cosign
4. For TUF, either mock out a TUF repo that will include a trusted root file, or just generate a trusted root file. Given all clients support providing trust root material via a file and don't require a full TUF repo, just generating a trusted root file seems like a good approach.

Once this is done, we can update Cosign's e2e tests to use this script.

It should be possible for this script to be runnable locally as well, so the GitHub Action script should do as little as possible, and we should provide a shell script for orchestration of the services (e.g. generate key material, spin up services, output trusted root).

Let me know if you have more questions.

jku:
I think we want this script to (at least optionally, but possibly by default) bring up both rekor v1 and rekor-tiles: this will most accurately match the planned initial infrastructure setup.

jku:
Maybe a dumb question but is the idea to (eventually) get the  compose file(s) to use published container images whenever possible?

haydentherapper:
> IMO the tuf scripts are not a positive for much the same reasons as the k8s config: they have diverged from the actual TUF setup that clients test against. Additionally their existence gives the impression that they are a reasonable way to run a production system when they really are not.
> I think option 1 the realistic one unless someone with interest and ideas for option 2 shows up...

Agreed, this is a risk of the current scaffolding repo, which is why ideally I'd like to deprecate most of what is here currently. Scaffolding is not representative of a real-world production environment, for TUF or the Sigstore services.

I mentioned the Scaffolding TUF tooling not for having it as a service but for leveraging relevant functions for transforming root key material into a trusted_root file, same as trtool.

> I think we want this script to (at least optionally, but possibly by default) bring up both rekor v1 and rekor-tiles, not one or the other: this will most accurately match the planned initial infrastructure setup.

Yes, though I think this is straightforward to support both, just have a flag for v1 or v2 (default).

> Maybe a dumb question but is the idea to (eventually) get the compose file(s) to use published container images whenever possible?

Yes, but not as a priority, though this seems straightforward. Like running V1 vs V2, control this with a flag on a shell script, e.g. with `--head`, pull from head, `--latest`, use a published container.

------------

I want to make sure we are all in alignment on what the purpose of this task is, which is to get a similar setup to what we have in Cosign e2e tests, which is a) spinning up the 3 core services, b) a fake OIDC provider trusted by Fulcio, c) a trusted root file.

Portability is not critical. If this runs on Linux, that's sufficient for CI testing, and we can iron out issues for local development iteratively. Also, as I said in the design doc, use a shell script to orchestrate standing up services and creating a trusted root if this cannot all be done via a single compose file.

After that, we can provide a GitHub Action reusable workflow for client developers, to effectively replace https://github.com/sigstore/scaffolding/tree/main/actions/setup.

.github/workflows/test-portable.yaml
ramonpetgrave64: Caching works now, but it's somehow slower than not using the cache, so i'm disabling it, and users can enable it with an option.

ramonpetgrave64:
@haydentherapper @jku It makes sense to reuse cosign's fakeoidc, so we'll have to prepare a Docekrfile or a docker-compose.yml file for it.

 - https://github.com/sigstore/cosign/tree/37bae90768f66c930b5630d0f570778141878737/test/fakeoidc

I agree the TUF things can probably come later, and primarily replace all the service standups that cosign's e2e tests do.

 - https://github.com/sigstore/cosign/blob/37bae90768f66c930b5630d0f570778141878737/test/e2e_test.sh

I think for now we can settle for this Makefile and Actions standing rekor and fulcio, and then the TSA and fakeoidc can come in followup PRs. wdyt?

jku:
> I think for now we can settle for this Makefile and Actions standing rekor and fulcio, and then the TSA and fakeoidc can come in followup PRs. wdyt?

 I do agree we need to focus this PR a bit ... however, the proposal doesn't get us much closer to testing any instance completely end-to-end (as in, a client could use this testing instance to sign and verify) and I think that would make sense as a goal. I would suggest two steps:

1. Goal for This PR: _"current cosign can be used with this test instance by just feeding it a trustedroot"_. The required missing things seem to be
   * fake-oidc 
   * trustedroot / signingconfig is not produced
2. For follow ups: _"test instance can be configured to support rekor v2"_. Missing things:
   * rekor v2 and rekor v1 are separate services that we want to run at the same time, not something we want to switch between
   * timestamp must be included

This order makes sense to me since we don't have rekorv2 capable clients anyway quite yet.  Reaching step 1 would mean cosign could consider using this already.

Things I believe we can drop from this PR for now to streamline:
* portability: I don't think this is something we should test for and it can make life a lot harder
* the action itself: Unsure how much easier this makes things but if the main planned users (cosign, rekor-tiles) plan to use this with the script, then let's not waste time polishing the action. Let's keep a test workflow though: we do want to see how reliable/slow this is on GHA

jku:
> Goal for This PR: _"current cosign can be used with this test instance by just feeding it a trustedroot"_. The required missing things seem to be
> * fake-oidc
> * trustedroot / signingconfig is not produced

I can start looking at the trustedroot part if you handle fake-oidc, does that sound good?

jku:
> * Goal for This PR: _"current cosign can be used with this test instance by just feeding it a trustedroot"_. The required missing things seem to be
>   
>   * fake-oidc
>   * trustedroot / signingconfig is not produced

oh, I did not realize the current Makefile actually installs rekor-tiles and not rekor. So in reality we only have fulcio running at this point for goal 1 :( maybe this proposal isn't good...

jku:
I've spent a little time looking at the trust root generation in https://github.com/sigstore/rekor-tiles/issues/150: I think I have decent solution now (although it requires a bug fix on sigstore-go that needs to be used in cosign...)

I would suggest TODO list for this PR is:
* Maybe remove the action for now? (up to you, I would simplify by removing)
* choose one of
  * Support both rekor v1 and rekor v2 (not as exclusive options, but as separate services), Support timestamp
  * Support rekor v1 only (not rekor-tiles as it now does)
* include the build-trusted-root script (I can provide a patch)
* Add a usage example in the README that show that this is relatively ergonomic to use " this is how to start a sigstore-in-a-box, this is how to run a client against it"

jku:
@ramonpetgrave64 I have a very early version of this that seems to work (with all services and trust root generation): it's full of hacks but it might be a way forward. let's discuss when you're back at office

ramonpetgrave64:
redoing in https://github.com/sigstore/scaffolding/pull/1544

PR #1606, opened 2025-06-17
Use service URLs and server name in trusted root

cmurphy:
Update cosign and use the new flags to set `uri` and `baseUrl` fields in trusted_root.json, and also to use the server name to generate the correct checkpoint verifier for Rekor v2.


cmurphy:
@haydentherapper @ramonpetgrave64

ramonpetgrave64 (approved):

haydentherapper (approved):

PR #1618, opened 2025-06-27
Parallelize service setup

cmurphy:
Speed up the setup-sigstore-env script by downloading and starting all four services in parallel.

This reduces the duration of the setup-sigstore-env action from about 10-11 minutes to about 9 minutes.

### DIFF
--- a/actions/setup-sigstore-env/run-containers.sh
+++ b/actions/setup-sigstore-env/run-containers.sh
@@ -52,32 +52,26 @@ OWNER_REPOS=(
   "$TIMESTAMP_AUTHORITY_REPO"
   "$REKOR_TILES_REPO"
 )
+procs=${#OWNER_REPOS[@]}
 for owner_repo in "${OWNER_REPOS[@]}"; do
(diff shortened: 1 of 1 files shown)


cmurphy:
@haydentherapper @ramonpetgrave64

ramonpetgrave64 (approved):
It's good to have the speedup.

One downside is that the logs can now be harder to debug through if we ever have errors, which would not happen when launched serially.

Some spanner log messages are now very close to, or mixed with, some trillian log messages

- https://github.com/sigstore/scaffolding/actions/runs/15937839666/job/44961144498#step:3:952
- https://github.com/sigstore/scaffolding/actions/runs/15937839666/job/44961144498#step:3:965

It's not too bad, but I suggest we make the parallel runs optional in the Action.

haydentherapper (commented):
re: logs, while this does interleave log messages from services, each log message should denote what service it comes from, which should be sufficient for debugging purposes.

haydentherapper (approved):

PR #1662, opened 2025-08-12
macOS support for hermetic environment

haydentherapper:
The primary change is handling the network connection between Fulcio and the fakeoidc service. On macOS, containers run in VMs and so they do not share a network by default. On Linux, containers can connect to one another through shared IP addresses without mucking around in /etc/hosts. If we made a request to the fakeoidc service without hardcoding the issuer, then the issuer would be 'localhost', and Fulcio couldn't connect to it since it's running in a VM. The workaround is to simply hardcode the container name as the issuer, and as before, join the fakeoidc container to Fulcio's network.

Also updated Cosign's container and did a little comment cleanup.


.github/workflows/cloud-sql-proxy-update.yml
```diff
@@ -8,7 +8,7 @@ permissions: {}
 
 jobs:
   update:
-  - name: Update cloud-sql-proxy
+    name: Update cloud-sql-proxy
```
haydentherapper, line 11: Yaml syntax is fun....

.github/workflows/test-setup-sigstore-env.yml
```diff
@@ -36,18 +36,17 @@ jobs:
             --new-bundle-format=true \
             --rekor-url http://localhost:3000 \
             --fulcio-url http://localhost:5555 \
-            --oidc-issuer ${STEPS_SETUP_SIGSTORE_ENV_OUTPUTS_OIDC_URL} \
```
haydentherapper, line 39: This flag was unnecessary

actions/setup-sigstore-env/run-containers.sh
```diff
@@ -60,23 +61,23 @@ for owner_repo in "${OWNER_REPOS[@]}"; do
     else
         echo "'cd $repo && git pull'"
     fi
-done | xargs -P "$procs" -l1 bash -c
+done | xargs -P "$procs" -L1 bash -c
```
haydentherapper, line 64: `-l` -> `-L` for POSIX compatibility

actions/setup-sigstore-env/Dockerfile.cosign
```diff
@@ -12,8 +12,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# We use this dockerfile for cosign so that Github's dependabot can autoupdate.
```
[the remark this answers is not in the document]
haydentherapper: Added back, I thought it was unnecessary but happy to leave it in!

bobcallaway (approved):

PR #1663, opened 2025-08-14
Get services from signing config for prober

aaronlew02:
Partially addresses sigstore/rekor-tiles#46

#### Summary

This change updates the prober to get its Fulcio, Rekor, and timestamping services from a signing config (TUF signing config v0.2 by default, unless overridden by a flag) instead of URL flags.

### DIFF
--- a/.github/workflows/prober-test.yml
+++ b/.github/workflows/prober-test.yml
@@ -35,6 +35,16 @@ jobs:
           go-version-file: 'go.mod'
           check-latest: true
 
-      - name: Prober test
+      - name: Build prober test
+        run: go build ./cmd/prober
+
+      - name: Get test OIDC token
+        uses: sigstore-conformance/extremely-dangerous-public-oidc-beacon@main
+
+      - name: Export OIDC token
+        run: |
+          echo "SIGSTORE_ID_TOKEN=$(cat ./oidc-token.txt)" >> $GITHUB_ENV
+
+      - name: Run prober test
         id: prober-test
-        run: go run ./cmd/prober --one-time --write-prober --logStyle dev -tsa-url https://timestamp.sigstage.dev
+        run: ./prober --one-time --write-prober --logStyle dev
--- a/cmd/prober/endpoints.go
+++ b/cmd/prober/endpoints.go
@@ -72,12 +72,7 @@ var FulcioEndpoints = []ReadProberCheck{
 var tsReq, _ = base64.StdEncoding.DecodeString("ME8CAQEwMTANBglghkgBZQMEAgEFAAQg6lDWJ0V9nVEPspa3bDKpG71ef/PswFWOcCjDxLBpe0cCFHsLm2h6a5KYc06qrKCtCIDhwZdmAQH/")
 var TSAEndpoints = []ReadProberCheck{
 	{
-		Endpoint: "/api/v1/timestamp/certchain",
-		Method:   GET,
-		Accept:   "application/pem-certificate-chain",
-	},
-	{
-		Endpoint:    "/api/v1/timestamp",
+		Endpoint:    "", // TSA endpoints are non-standard and come from the signing config
 		Method:      POST,
 		Accept:      "application/timestamp-reply",
 		ContentType: "application/timestamp-query",
--- a/cmd/prober/prober.go
+++ b/cmd/prober/prober.go
@@ -37,6 +37,10 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 	fulciopb "github.com/sigstore/fulcio/pkg/generated/protobuf"
+	prototrustroot "github.com/sigstore/protobuf-specs/gen/pb-go/trustroot/v1"
+	"github.com/sigstore/sigstore-go/pkg/root"
+	"github.com/sigstore/sigstore-go/pkg/sign"
+	"github.com/sigstore/sigstore-go/pkg/tuf"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/credentials"
 	"google.golang.org/grpc/status"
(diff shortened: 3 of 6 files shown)


.github/workflows/prober-test.yml
```diff
@@ -35,6 +35,13 @@ jobs:
           go-version-file: 'go.mod'
           check-latest: true
 
+      - name: Get test OIDC token
+        uses: sigstore-conformance/extremely-dangerous-public-oidc-beacon@main
+
+      - name: Export OIDC token
+        run: |
+          echo "SIGSTORE_ID_TOKEN=$(cat ./oidc-token.txt)" >> $GITHUB_ENV
```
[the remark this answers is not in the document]
aaronlew02: Changing the order of the workflow to reduce latency is a great idea.

I don't know that this workflow is used in any such critical situations, or even anywhere outside of this CI/CD pipeline.

actions/setup-sigstore-env/run-containers.sh
```diff
@@ -74,10 +74,13 @@ done | xargs -P "$procs" -L1 bash -c
 # network to enable Fulcio to reach it for token verification.
 docker network inspect fulcio_default | grep fakeoidc || docker network connect --alias fakeoidc fulcio_default fakeoidc || return
 export TSA_URL="http://localhost:3004"
+export FULCIO_CERT_COUNT=2
```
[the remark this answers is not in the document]
aaronlew02: The hermetic environment's Fulcio service returns two (not three) certificates, so the expected certificate count must be configurable. This will be extracted into a separate PR dedicated to local testing.

cmd/prober/prober.go
```diff
@@ -104,45 +107,119 @@ func encodeLevel() zapcore.LevelEncoder {
...
+		log.Fatal("Must specify both --signing-config and --trusted-root, or neither")
+	}
+
+	rekorV1Services, err = root.SelectServices(signingConfig.RekorLogURLs(), root.ServiceConfiguration{Selector: prototrustroot.ServiceSelector_ALL}, []uint32{1}, time.Now())
+	if err != nil {
+		log.Fatal("Failed to select Rekor services: ", err)
+	}
+	rekorV1URLs = mapServicesToURLs(rekorV1Services)
+
+	fulcioService, err = root.SelectService(signingConfig.FulcioCertificateAuthorityURLs(), []uint32{1}, time.Now())
```
haydentherapper, line 201: You can also use https://github.com/sigstore/sigstore-go/blob/070130677531d3e49c84866bee26d43ea710d2ac/pkg/sign/certificate.go#L61 to specify the version, I've exported variables for the supported versions for each service.

cmd/prober/prober.go
```diff
@@ -104,45 +107,119 @@ func encodeLevel() zapcore.LevelEncoder {
...
+		}
+	default:
+		log.Fatal("Must specify both --signing-config and --trusted-root, or neither")
+	}
+
+	rekorV1Services, err = root.SelectServices(signingConfig.RekorLogURLs(), root.ServiceConfiguration{Selector: prototrustroot.ServiceSelector_ALL}, []uint32{1}, time.Now())
+	if err != nil {
+		log.Fatal("Failed to select Rekor services: ", err)
+	}
+	rekorV1URLs = mapServicesToURLs(rekorV1Services)
```
haydentherapper, line 199: Can we make these local variables rather than global, now that we don't have the flag setting them?

cmd/prober/endpoints.go
```diff
@@ -40,10 +40,6 @@ var ShardlessRekorEndpoints = []ReadProberCheck{
 	}, {
 		Endpoint: "/api/v1/log",
 		Method:   GET,
-	}, {
```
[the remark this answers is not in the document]
haydentherapper: Ah I forget, we talked about this, that this test fails for logs that just spun up.

cmd/prober/prober.go
```diff
@@ -349,72 +457,73 @@ func httpRequest(host string, r ReadProberCheck) (*retryablehttp.Request, error)
 	return req, nil
 }
 
+func mapServicesToURLs(services []root.Service) []string {
```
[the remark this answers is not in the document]
aaronlew02: I have decided to take your approach and in #1719 will create a service slice for Rekor v2 in its prober if a URL is provided by flag.

cmd/prober/prober.go
aaronlew02: URL variables have been removed.

cmd/prober/prober.go
```diff
@@ -104,44 +108,118 @@ func encodeLevel() zapcore.LevelEncoder {
...
-	addr           string
-	rekorURL       string
-	fulcioURL      string
-	fulcioGrpcURL  string
-	tsaURL         string
 	oneTime        bool
 	runWriteProber bool
-	versionInfo    version.Info
+
+	rekorV1Services []root.Service
```
haydentherapper, line 125: Can we declare these variables and `err` when we use them? Let's only declare global variables for the flag values.

cmd/prober/prober.go
```diff
@@ -104,44 +108,118 @@ func encodeLevel() zapcore.LevelEncoder {
...
+
 	flag.IntVar(&frequency, "frequency", 10, "How often to run probers (in seconds)")
-	flag.StringVar(&logStyle, "logStyle", "prod", "log style to use (dev or prod)")
+	flag.StringVar(&logStyle, "logStyle", "prod", "Log style to use (dev or prod)")
 	flag.StringVar(&addr, "addr", ":8080", "Port to expose prometheus to")
+	flag.IntVar(&grpcPort, "grpc-port", 5554, "Port for Fulcio gRPC endpoint (only if configured)")
+
+	flag.UintVar(&retries, "retry", 4, "Maximum number of retries before marking HTTP request as failed")
+	flag.BoolVar(&oneTime, "one-time", false, "Whether to run only one time and exit")
+	flag.BoolVar(&runWriteProber, "write-prober", false, " [Kubernetes only] run the probers for the write endpoints")
```
haydentherapper, line 153: Is this still accurate to only run on K8s? It now works locally and against the hermetic environment, correct?

cmd/prober/prober.go
```diff
@@ -104,44 +108,118 @@ func encodeLevel() zapcore.LevelEncoder {
...
 
-	flag.StringVar(&rekorURL, "rekor-url", "https://rekor.sigstore.dev", "Set to the Rekor URL to run probers against")
-	flag.StringVar(&fulcioURL, "fulcio-url", "https://fulcio.sigstore.dev", "Set to the Fulcio URL to run probers against")
-	flag.StringVar(&fulcioGrpcURL, "fulcio-grpc-url", "fulcio.sigstore.dev", "Set to the Fulcio GRPC URL to run probers against")
+	flag.StringVar(&rekorRequestsJSON, "rekor-requests", "[]", "Additional rekor requests (JSON array)")
```
haydentherapper, line 155: nit, can we rename this variable to include `v1` to distinguish it from the upcoming change to support v1?

haydentherapper (approved):
Just one minor comment, this looks great!

haydentherapper (approved):
🥳

cmd/prober/prober.go
```diff
@@ -104,42 +108,46 @@ func encodeLevel() zapcore.LevelEncoder {
 }
 
 var (
-	logStyle       string
-	frequency      int
+	scPath      string
+	trPath      string
+	trustedRoot *root.TrustedRoot
```
cmurphy, line 113: This also doesn't need to be global. These global variables are meant to be assigned to command line flag values, but otherwise it's best to instantiate the variable locally and pass it as an argument to the functions that need it.

cmd/prober/prober.go
```diff
@@ -104,42 +108,46 @@ func encodeLevel() zapcore.LevelEncoder {
...
-	rekorURL       string
-	fulcioURL      string
-	fulcioGrpcURL  string
-	tsaURL         string
 	oneTime        bool
 	runWriteProber bool
-	versionInfo    version.Info
+
+	rekorV1RequestsJSON string
+	fulcioRequestsJSON  string
```
cmurphy, line 126: These also don't need to be moved into globals, they aren't used anywhere outside of `init()` so I would leave them where they were

PR #1793, opened 2025-11-19
Replace ct_server with TesseraCT in setup

cmurphy:
In the setup scaffolding workflow, update Fulcio to use the Static CT
log TesseraCT instead of the Trillian-based ct_server.

This change avoids retrieving the Fulcio roots from the rootCert
endpoint, instead favoring using the known externally generated root,
which closer simulates the mechanism for distributing key material out
of band via TUF. This means the createctconfig job is not needed for
this CT log. The createtree job is also no longer needed because there
is no Trillian tree to manage.

Relates to https://github.com/sigstore/rekor-tiles/issues/73


.github/workflows/add-remove-new-fulcio.yaml
haydentherapper: Is this test still meaningful or can we just remove it? Testing that the CT log supports rotating trusted roots is testing the functionality of the CT server rather than the scaffolding setup. Also, now that we're recommending a manual process of applying a secret, we don't need to test that our custom jobs or scripts work.

Also means we can remove the verifyfulcio script

haydentherapper (commented):
Can we leave a TODO or warning in the README that the [ctlog](https://github.com/sigstore/scaffolding/blob/main/README.md#ctlog) section is now out of date?

cmurphy:
> Can we leave a TODO or warning in the README that the ctlog section is now out of date?

I went ahead and updated it.

haydentherapper (approved):
Looks great, love seeing a PR with lots of red! Up to you if you want to merge the other PR first and rebase or rebase this PR onto main.

haydentherapper (approved):

PR #1820, opened 2025-11-25
group dependabot fixes

bobcallaway:


loosebazooka:
shoot bob, I just realized we should've held off on this before #1814 so that the updates can look into the new paths

PR #1946, opened 2026-07-06
workflows: Allow insecure registries in tests

### LINKED ISSUE #1945: fulcio-rekor-kind.yml is broken
All runs of fulcio-rekor-kind.yml  are failing in "Create sample image" step:

```
go: creating new go.mod: module example.com/demo
2026/07/04 05:28:50 Using base cgr.dev/chainguard/static:latest@sha256:77d8b8925dc27970ec2f48243f44c7a260d52c49cd778288e4ee97566e0cb75b for example.com/demo
2026/07/04 05:28:51 current folder is not a git repository. Git info will not be available
2026/07/04 05:28:51 Building example.com/demo for linux/amd64
2026/07/04 05:29:02 Publishing registry.local:5000/knative/demo:latest
Error: failed to publish images: error publishing ko://example.com/demo: writing sbom: Get "https://registry.local:5000/v2/": http: server gave HTTP response to HTTPS client
Error: Process completed with exit code 1.
```

My guess is that the new ko release (0.19) is breaking this

jku:
setup-ko installs newest ko. ko 0.19 started bringing in an upgraded go-containerregistry which seems to restrict non-HTTPS use.

* fulcio-rekor-kind: create a "ko" wrapper that allows insecure registries (because there are loads of ko calls being made via Makefile). This script is AI generated
* The other two workflows only make a single ko call: modify the call sites

Fixes #1945 but I'd appreciate if someone more familiar with this setup would decide if this is the right fix -- maybe we do want to change all the places where ko is called to use "--insecure-registry" instead?

### DIFF
--- a/.github/workflows/fulcio-rekor-kind.yaml
+++ b/.github/workflows/fulcio-rekor-kind.yaml
@@ -77,6 +77,36 @@ jobs:
 
     - uses: ko-build/setup-ko@d006021bd0c28d1ce33a07e7943d48b079944c8d # v0.9
 
+    - name: Set up ko wrapper for insecure registry
+      run: |
+        mkdir -p ${{ github.workspace }}/bin
+        cat <<'EOF' > ${{ github.workspace }}/bin/ko
+        #!/usr/bin/env bash
+        MY_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+        for dir in ${PATH//:/ }; do
+          if [ "$dir" != "$MY_DIR" ] && [ -x "$dir/ko" ]; then
+            REAL_KO="$dir/ko"
+            break
+          fi
+        done
+        if [ -z "$REAL_KO" ]; then
+          echo "Real ko not found" >&2
+          exit 1
+        fi
+        cmd="$1"
+        shift
+        case "$cmd" in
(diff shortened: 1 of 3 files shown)


Hayden-IO:
Could we do the opposite and make registry calls over HTTPS? Not sure if that would be any cleaner though, I imagine we'd either need to generate and trust a root certificate or trust one generated by ko (I assume it would, given HTTPS is now the default).

jku:
> Could we do the opposite and make registry calls over HTTPS? Not sure if that would be any cleaner though, I imagine we'd either need to generate and trust a root certificate or trust one generated by ko (I assume it would, given HTTPS is now the default).

I think that might be a significant effort: It looks like the registry setup is done by either setup-kind.sh or the setup-kind action -- neither seems handle setting up HTTPS and the latter doesn't have a way to trust a cert even if we did have a CA of our own

Hayden-IO (approved):
Let's get this in so we can unblock CI.

cmurphy (approved):
I don't love the wrapper script - what if we added an environment variable to the Makefile to optionally include --insecure in the ko commands? But this is fine to unblock CI.

Hayden-IO:
Going to merge since this is blocking CI, let's follow up on the cleanup for the script.

PR #1952, opened 2026-07-07
Support rekor-tiles in release setup script, fix 404 check

Hayden-IO:
The Rekor release config has been renamed, so in the event that release-rekor is missing, switch to release-rekor-tiles.

Also when testing the script, I saw that the check for the Trillian config wasn't being skipped as expected. Fixes suggested by Gemini.

### DIFF
--- a/hack/setup-scaffolding-from-release.sh
+++ b/hack/setup-scaffolding-from-release.sh
@@ -79,7 +79,7 @@ trap cleanup EXIT
 # Install Trillian if it is part of this release and wait for it to come up
 echo '::group:: Install Trillian'
 has_trillian=1
-if curl -s -i "https://github.com/sigstore/scaffolding/releases/download/${RELEASE_VERSION}/release-trillian.yaml" | grep -q '404' ; then
+if curl -s -i "https://github.com/sigstore/scaffolding/releases/download/${RELEASE_VERSION}/release-trillian.yaml" | grep -q 'HTTP/.* 404' ; then
   has_trillian=0
 fi
 if [ "$has_trillian" == "1" ]; then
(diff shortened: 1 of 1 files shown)


hack/setup-scaffolding-from-release.sh
```diff
@@ -78,15 +79,19 @@ trap cleanup EXIT
 # Install Trillian if it is part of this release and wait for it to come up
 echo '::group:: Install Trillian'
 has_trillian=1
-if curl -s -i "https://github.com/sigstore/scaffolding/releases/download/${RELEASE_VERSION}/release-trillian.yaml" | grep 'HTTP/2 404' ; then
+if curl -s -i "https://github.com/sigstore/scaffolding/releases/download/${RELEASE_VERSION}/release-trillian.yaml" | grep -q '404' ; then
```
cmurphy:
```suggestion
if curl -s -i "https://github.com/sigstore/scaffolding/releases/download/${RELEASE_VERSION}/release-trillian.yaml" | grep -q 'HTTP/.* 404' ; then
```
I worry that just grepping for 404 will find some innocuous numbers in one of the headers that will give a false positive.

hack/setup-scaffolding-from-release.sh
```diff
@@ -100,6 +105,9 @@ cleanup_rekor() {
     rm "${rekordir}/pub.pem" "${rekordir}/key.pem"
 }
 cleanup_cmd="cleanup_rekor"
+if curl -s -i "${REKOR}" | grep -q '404' ; then
```
cmurphy: Same here

hack/setup-scaffolding-from-release.sh
Hayden-IO: Added

cmurphy (approved):

PR #1964, opened 2026-08-14
pull containers from GHCR instead of GCR

bobcallaway:
no need to use GCR

### DIFF
--- a/.github/workflows/test-release.yaml
+++ b/.github/workflows/test-release.yaml
@@ -70,7 +70,9 @@ jobs:
         version: "1.18.x"
         serving-features: >
           {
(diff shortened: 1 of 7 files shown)


Hayden-IO (dismissed):

Hayden-IO:
Oh you might be running into the same failures from when I explored updating k8s previously

bobcallaway:
> Oh you might be running into the same failures from when I explored updating k8s previously

There’s still a chicken/egg problem with the way these tests are structured.

Hayden-IO (dismissed):

Hayden-IO (dismissed):

